PR #10, opened 2024-05-08
Fix calling convention on putchar (silgen_name -> cdecl)

kubamracek:
This should make the example projects build agains the current nightly toolchains. The compiler change that starting causing the mismatch is this one: https://github.com/apple/swift/pull/72973


rauhul (approved):
good catch!

PR #27, opened 2024-06-29
update gitignore file

erhankur:
When cmake configures the project, "dependencies.lock" file will be generated automatically. So no need to keep it in the repo.

More context related to component manager https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/tools/idf-component-manager.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,6 @@
 */.build
 */build
 */sdkconfig
+*/sdkconfig.old
 */managed_components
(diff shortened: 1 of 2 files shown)


rauhul:
I think this looks good, @kubamracek can you confirm that its ok to remove the `.lock` file?

rauhul (approved):
Thanks for the contribution!

PR #29, opened 2024-07-08
Use new swift-mmio naming @RegisterBlock, fix build of STM32 example

kubamracek:
`@RegisterBank` has been changed to `@RegisterBlock` in swift-mmio, let's reflect that.


rauhul (approved):
Thank you!

PR #30, opened 2024-07-08
[stm32-neopixel] Use new @RegisterBlock name, remove warning about swiftpm

kubamracek:
Same as https://github.com/apple/swift-embedded-examples/pull/29, but for `stm32-neopixel`.


rauhul (approved):
thank you!

Issue #20, opened 2024-06-23
Instructions don't mention installing python dependencies


CraigSiemens:
When building `pico-blink` for the first time, it produces the following error.

```
ModuleNotFoundError: No module named 'macholib'
```

After looking through the project I found the `requirements.txt` file and ran the following to fix the issue.

```shell
pip3 install -r ../Tools/requirements.txt
```

rauhul:
Would you be interested in PR a documentation fix for this?

mredig:
I actually just updated the docs myself with this and one or two more documentation clarifications. Is there a more formal process like the larger Swift project, or should I just open a PR?

rauhul:
Just open a PR :)

PR #39, opened 2024-07-30
Update pico readmes with additional clarity

mredig:
I ran into a few issues that weren't easily googleable and therefore required a decent level of troubleshooting to discover, so I updated the docs to help others overcome these hurdles.

### DIFF
--- a/pico-blink/README.md
+++ b/pico-blink/README.md
@@ -23,9 +23,10 @@ pip3 install -r requirements.txt
 ```
 
 - Build and copy the program in the UF2 format to the Mass Storage device to trigger flashing the program into memory (after which the device will reboot and run the firmware):
+  - (determining your toolchain identifier is step 5 in this [matter tutorial](https://apple.github.io/swift-matter-examples/tutorials/swiftmatterexamples/setup-macos/))
 ``` console
 $ cd pico-blink
-$ TOOLCHAINS='<toolchain-name>' ./build.sh # toolchain-name is typically `swift`
+$ TOOLCHAINS='<toolchain-identifier>' ./build.sh
 $ cp .build/blink.uf2 /Volumes/RP2040
 ```
 - The green LED should now be blinking in a pattern.
--- a/pico-blink/README.md
+++ b/pico-blink/README.md
@@ -10,10 +10,22 @@
 
 - Connect the Pico board via a USB cable to your Mac, and make sure it's in the USB Mass Storage firmware upload mode (either hold the BOOTSEL button while plugging the board, or make sure your Flash memory doesn't contain any valid firmware).
 - Make sure you have a recent nightly Swift toolchain that has Embedded Swift support.
+- Before building, you need to install the python dependencies (ideally in a virtual environment):
+
+```console
+cd swift-embedded-examples
+mkdir pyenv
+cd pyenv
+python3 -m venv .
+source bin/activate
+cd ../Tools
+pip3 install -r requirements.txt
+```
+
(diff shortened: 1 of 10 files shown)


mredig:
Resolves Issue #20

pico-blink/README.md
```diff
@@ -10,10 +10,22 @@
...
+source bin/activate
+cd ../Tools
+pip3 install -r requirements.txt
+```
+
 - Build and copy the program in the UF2 format to the Mass Storage device to trigger flashing the program into memory (after which the device will reboot and run the firmware):
 ``` console
 $ cd pico-blink
-$ TOOLCHAINS='<toolchain-name>' ./build.sh
+$ TOOLCHAINS='<toolchain-name>' ./build.sh # toolchain-name is typically `swift`
```
kubamracek, line 28: I don't think is actually true. I have multiple toolchains installed and if I set TOOLCHAINS=swift, I do *not* get the latest one chosen.

But point taken that the instructions are not really clear on what the TOOLCHAINS variable should be set to...

rauhul: I dont think we want to suggest `swift` as the toolchain name because when a user has multiple toolchains installed its very very unclear which one is getting used.

mredig: I agree! How would I fix this instruction? It's the only thing that I could find that worked for *my* situation, but it wasn't clear where I was supposed to source a value from. (I had also tried a path to the toolchain via a ChatGPT suggestion amongst several other attempts, but none of those worked)

rauhul: I described the instructions here: https://apple.github.io/swift-matter-examples/tutorials/swiftmatterexamples/setup-macos

But its annoying we keep repeating these instructions across tons of examples/tutorials

mredig: I added a link to those instructions - thoughts on that?

mredig: I also changed `toolchain-name` to `toolchain-identifier` since that seems to more accurately reflect what it is and what it's referred to as in the given instructions. (not to mention it would have clarified a lot of my own confusion when I was looking for a `name` and not an `identifier` - I think I even glanced into the plist and mentally discarded the id since that's not what was requested)

rauhul: > I added a link to those instructions - thoughts on that?

It's definitely an option and might be ok for now. @cmcgee1024 do you have thoughts on where we can host more details "how to use a custom toolchain" instructions? I really do not want to maintain this information in every embedded examples repo.

mredig: If you're given freedom to create your own tutorials like the matter one, I think that a tutorial on sourcing, using, identifying, etc on toolchains might be worth it on its own.

rauhul: We could certainly create such a tutorial in this repo, but I would prefer we have something hosted on Swift.org

mredig: I just meant in the same style as the matter one, not really intending a specific location for it to go. Wherever the home, links could send people there.

rauhul:
@mredig I've rebased and pushed some changes to the approach you had taken. Since we have multiple pico examples and the STM examples also need macholib I've elevated the python instructions to the top level readme. 

Let me know what you think!

rauhul:
Im gonna merge this as is, feel free to comment and/or open a pr with changes if desired

mredig:
> Im gonna merge this as is, feel free to comment and/or open a pr with changes if desired

Ah I'm sorry! I missed the notification. It all looks good though! :D

PR #51, opened 2024-08-19
Add an RP2350 example project

rauhul:


.gitignore
```diff
@@ -6,3 +6,4 @@
 */sdkconfig.old
 */managed_components
 */dependencies.lock
+nrfx-blink-sdk
```
kubamracek, line 9: this probably shouldn't be here, right?

kubamracek (approved):

PR #57, opened 2024-08-23
Update the Pico2 example for more boards

rauhul:
Updates the Pico2 example to use the patched RP2350 SVD file from the rust community and updates to a newer version of swift-mmio with support for emitted "dimensioned" SVD elements as arrays.


rauhul:
I need to actually try this on my device before merging!

rauhul:
verified it locally

Issue #49, opened 2024-08-16
[Question] `undefined reference to '_swift_exceptionPersonality'` while building a project for pi pico


Helloyunho:
I'm trying to move my original C++ code to Swift. Since my project uses TinyUSB for HID keyboard control, I decided to copy paste some codes from `pico-blink-sdk`.

Here's what `CMakeLists.txt` looks like:
```cmake
cmake_minimum_required(VERSION 3.13)

include(pico_sdk_import.cmake)

project(PROJECT_NAME_HERE_project C CXX ASM)
set(CMAKE_C_STANDARD 11)
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
pico_sdk_init()

if(APPLE)
execute_process(COMMAND xcrun -f swiftc OUTPUT_VARIABLE SWIFTC OUTPUT_STRIP_TRAILING_WHITESPACE)
else()
execute_process(COMMAND which swiftc OUTPUT_VARIABLE SWIFTC OUTPUT_STRIP_TRAILING_WHITESPACE)
endif()

add_executable(PROJECT_NAME_HERE)

target_include_directories(PROJECT_NAME_HERE PUBLIC ${CMAKE_CURRENT_LIST_DIR})

target_link_libraries(PROJECT_NAME_HERE
    pico_stdlib
    pico_unique_id
    tinyusb_device
    tinyusb_board
    ${CMAKE_CURRENT_BINARY_DIR}/_swiftcode.o
)

add_custom_command(
    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/_swiftcode.o
    COMMAND
        ${SWIFTC}
        -target armv6m-none-none-eabi -Xcc -mfloat-abi=soft -Xcc -fshort-enums
        -Xfrontend -function-sections -enable-experimental-feature Embedded -wmo -parse-as-library
        -cxx-interoperability-mode=default
        -Xcc -std=c++17
        $$\( echo '$<TARGET_PROPERTY:PROJECT_NAME_HERE,INCLUDE_DIRECTORIES>' | tr '\;' '\\n' | sed -e 's/\\\(.*\\\)/-Xcc -I\\1/g' \)
        $$\( echo '${CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES}'             | tr ' '  '\\n' | sed -e 's/\\\(.*\\\)/-Xcc -I\\1/g' \)
[...]

PR #58, opened 2024-08-29
Pico-blink-sdk support for RP2350

iCMDdev:
Hi, 

I added support for running the [pico-sdk blink example](https://github.com/apple/swift-embedded-examples/tree/main/pico-blink) on RP2350 hardware (using the ARM cores, more on supporting RISC-V below). 

Previously, when trying to compile the example for the Pico 2, one may encounter an "invalid instruction" error thrown by `swiftc`. I realized this is due to the architecture differences between RP2040 and RP2350: the former is a Cortex-M30 device using Armv6-M, while the latter is a Cortex-M33 using Armv8-M (but the swift compiler doesn't support it, so armv7em is being used, just as in the [pico2-neopixel example (without the SDK)](https://github.com/apple/swift-embedded-examples/tree/a3b2df77ec406184136b40a246de0c6933119362/pico2-neopixel)). 

This pull request changes the project's `CMakeLists.txt` file to automatically account for the target architecture change, guiding `swiftc` to use the relevant target based on the `PICO_PLATFORM` CMake variable (which, if not set by the user, is automatically set based on `PICO_BOARD`). 

I've also added a target check for RISC-V (since the RP2350 has the option to run the code in RISC-V mode), but for me, compilation fails, showing some errors related to the pico-sdk. I also have some similar errors when compiling for the rp2040 with the new SDK (not exactly the issue described in #49, but changing to previous SDK versions temporarily fixes them).

Edit: the SDK-related RP2040 issue I mentioned is #59.


iCMDdev:
Thanks for reviewing and merging this! 

> Do you have any ideas on how we can organize the pico examples to be more clear about what devices are supported overall?

Hmm... I'm not sure. On one hand, Raspberry Pi themselves advertise the new RP2350 as being compatible with RP2040 software, so, when someone sees "Raspberry Pi Pico" they could assume it works on the new chips as well.

Perhaps a solution would be to use the terms "Raspberry Pi Pico series" / "Raspberry Pi Pico, Pico 2", "Pico (RP2040)" / "Pico (RP2040 only)", and "Pico 2 (RP2350)". What do you think?

PR #60, opened 2024-09-12
Add --pico-family parameter

sw-dev-1:
Compilation of the `pico-blink` example failed with the following error:

```
usage: macho2uf2.py [-h] --base-address BASE_ADDRESS --segments SEGMENTS --pico-family PICO_FAMILY input output
macho2uf2.py: error: the following arguments are required: --pico-family
```

Looks like the `--pico-family` parameter became mandatory.

This PR adds a `PICO_FAMILY` variable near the top of the build script, with a comment pointing out the two possible values (according to the source code of `macho2uf2.py`).
It then then sets `--pico-family` to this variable.

Maybe it makes sense to set up some kind of CI to catch these kind of build errors automatically.

### DIFF
--- a/pico-blink/build.sh
+++ b/pico-blink/build.sh
@@ -2,6 +2,9 @@
 
 set -vex
 
+# Either rp2040 or rp2350
+PICO_FAMILY=rp2040
+
 # Determine file paths
 REPOROOT=$(git rev-parse --show-toplevel)
 TOOLSROOT=$REPOROOT/Tools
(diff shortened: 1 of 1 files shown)


iCMDdev:
I think you could also change the target architecture from `armv6m` to `armv7m` for the based on this for RP2350 boards, even though the latter is backwards compatible with the former (and should be fine as long as the swift binary isn't being linked to armv7m libs, like it was the case in the Pico SDK examples).

Maybe the slightly improved instruction set will bring some tiny performance benefits? 😁

rauhul:
Sorry I missed this (was out sick) I think we can merge this as is!

PR #61, opened 2024-09-15
Add GitHub action to build the examples

sw-dev-1:
This shows a proof-of-concept for a simple continuous integration system, using GitHub actions.

A [single file](https://github.com/sw-dev-1/swift-embedded-examples/blob/feature/ci/.github/workflows/build-embedded-swift-examples.yml) is added, defining the steps to build the code.

The steps are:
 - Download a recent development snapshot toolchain.
 - Setup Python venv and install Python dependencies.
 - Build `pico-blink` example, as a starting point.

Example runs:
 - Successful run: To show a `blinky.uf2` binary for pico-blink is produced, I applied the fix from [pull request 60](https://github.com/apple/swift-embedded-examples/pull/60). The corresponding, successfull run can be found [here](https://github.com/sw-dev-1/swift-embedded-examples/actions/runs/10847351435/job/30102247053).
 - Unsuccessful run: Without the patch from PR 60, the pico-blink example does not compile. This can be seen [here](https://github.com/sw-dev-1/swift-embedded-examples/actions/runs/10869890839)

The action is run daily at 9 AM UTC, this can be set to any other time.

Steps for the other examples can be added later, in case this approach is liked.


iCMDdev:
That's pretty cool, though I think Apple usually uses their own CI (at least in a few swift repos). But if a CI will be adopted I think you could also add automatic releases of UF2 binaries.

For example, I did this in [this repo](https://github.com/iCMDdev/swift-rp2350-lvgl-st7789), though it could be improved with version tags.

iCMDdev:
Good news: https://forums.swift.org/t/ci-is-it-possible-to-add-ci-checks-with-github-actions/74662/3

rauhul:
So there are some caveats, 1) I need to get them enabled on this repo (should be easy, I hope). 2) we can't use macOS runners yet

sw-dev-1:
2) I'll try to change it to a Linux runner. For the changes, should I add additional commits or force push, for a cleaner git history?

rauhul:
We will squash on merge so feel free to take whichever approach you prefer in the PR.

sw-dev-1:
Using Linux runners is more complicated than anticipated.

I managed to use Linux to build the `pico-blink` example, but ran into the following problems, all related to building on Linux:

- xcrun is not available on Linux. This can be worked around in the `build.sh` of `pico-blink` by adding `if [ $(uname) = "Linux" ]` checks to not use xcrun.
- Apple libtool and GNU libtool (Linux) behave different. On macOS build, the `-static` flag of libtool is used to produce static libraries. However, GNU libtool 2.4.7 does not know the `-static` flag (at least on [Ubuntu 24.04](https://manpages.ubuntu.com/manpages/oracular/man1/libtool.1.html)). Older versions of Ubuntu do seem to have to `-static` flag, as can be seen [here](https://manpages.ubuntu.com/manpages/trusty/man1/libtool.1.html)

My impression is, building for Linux has hardly been tested, if at all.

Going further, I see two options:
- Reject this pull request. The action runs in my private fork, any issues that show up there have to reported manually.
- You have to figure out how to use macOS runners.

iCMDdev:
> Going further, I see two options:
> 
> * Reject this pull request. The action runs in my private fork, any issues that show up there have to reported manually.
> * You have to figure out how to use macOS runners.

Why not also fix the Linux build? Especially if we won't be able to use macOS runners.

Looking at `build.sh`, there's a few things that need to be adjusted for Linux:

- First, as you already pointed out, xcrun should not be used on Linux:
```sh
if [ $(uname) = "Darwin" ]
  SWIFT_EXEC=${SWIFT_EXEC:-`xcrun -f swift`}
  CLANG=${CLANG:-`xcrun -f clang`}
  PYTHON_EXEC=${PYTHON_EXEC:-`xcrun -f python3`}
elif [ $(uname) = "Linux" ]
  SWIFT_EXEC=${SWIFT_EXEC:-`which swift`}
  CLANG=${SWIFT_EXEC:-`which clang`}
  PYTHON_EXEC=${PYTHON_EXEC:-`which python3`}
fi
```

- But also, I believe this might not be enough, and I suppose this is why we're seeing the libtool error. [Notice that `armv6m-apple-none-macho` is still used throughout the script, even though this is running on Linux.](https://github.com/apple/swift-embedded-examples/blob/22f2c981c524416fd4cf89ef84140a81cf2748b1/pico-blink/build.sh#L16) I suppose we might need to change these to `armv6m-none-none-eabi` and change the compilation / linking steps accordingly, perhaps?

sw-dev-1:
> Why not also fix the Linux build? 

I'm not willing to do it. If somebody else wants to, please go ahead.

iCMDdev:
> > Why not also fix the Linux build? 
> 
> 
> 
> I'm not willing to do it. If somebody else wants to, please go ahead.

I'm willing to do it but I guess I'll need to document myself about this and find the time for it. I'm not sure how busy I'll be in the near future, but we'll see :)

Maybe some help / tips from others would be helpful!

sw-dev-1:
See https://github.com/sw-dev-1/swift-embedded-examples/actions for the build status.

PR #67, opened 2024-10-26
Raspberry Pi 4B & 5 baremetal blink example

iCMDdev:
Currently just a draft for a baremetal Raspberry Pi 4B 64-bit blink example (`-target aarch64-none-none-elf`). I'll add the code soon - I'm making this draft public now to receive suggestions & feedback as I work on this example.

I already got Embedded Swift working on a 4B (blinking a led with Swift-MMIO), but I need to do some cleanup & better organize the code & build process. I'll upload it here after I do that (it will be in a Swift Package format). 

I'm currently using `llvm-objcopy` as a last step to convert the resulted `ELF` file, with something like `$(LLVMPATH)/llvm-objcopy -O binary kernel8.elf kernel8.img`. From what I've seen, there doesn't seem to be any viable alternative, so users will need to install llvm with `brew`, since `llvm-objcopy` is not installed by default.


iCMDdev:
This doesn't need to be necesarily specific to the 4B, I might as well adapt it to the 5B. Some register values might be different, and also, it seems that the 5's ACT (green) LED is not directly connected to the GPIO, making it a bit harder to blink it directly. 

Still, one could blink another LED connected directly to the GPIO and changing the "pin value" (currently hardcoded in Swift MMIO registers in my example...)

iCMDdev:
That was faster than expected,  I think it's ready for review now!

iCMDdev:
`boot.S` and `link.ld` are borrowed from [this bare-metal code example on Raspberry Pi](https://github.com/babbleberry/rpi4-osdev/tree/master/part4-miniuart) (CC0-licensed).

iCMDdev:
I also got this working on a Pi 5! I managed to gather the relevant register details (adresses & necessary bit values).
Now, I'm wondering whether these should be 2 separate examples, or if I should merge them into one (and use compiler directives to specify which the board).

rauhul:
I think the example makes sense to add. However Im trying to get CI setup for this repo before adding additional examples, so avoid hosting broken code!

iCMDdev:
> However Im trying to get CI setup for this repo before adding additional examples

Make sense, alright! 

> so avoid hosting broken code!

I'm not sure what you mean by that, do you mean that this example is broken somewhere? I was able to build & run it just fine, but maybe that's on my machine only.

Either way, as you suggested, we should popstone this example and come back to it once we get CI integration. That's a good idea.

rauhul:
I mean the examples silently breaking due to changes in the swift compiler

iCMDdev:
> I mean the examples silently breaking due to changes in the swift compiler or other example dependencies, e.g. the SDKs where we depend on a branch.

Oh, right, yeah. Great idea.

rauhul:
@iCMDdev we now have some basic CI for the examples (excluding the current ones that only build for `macho`) Can you try to extend this CI to the new content in this PR?

iCMDdev:
Sure! I think I'll do this over the weekend.

iCMDdev:
I added the CI example, it should be working now. I'll now work on fixing the lint errors.

iCMDdev:
Seems to be fine now! I fixed the styling issues as well.

iCMDdev:
Thanks for running the checks! I’m not sure how I missed installing `make`, I know I had it in the script…  Maybe I did the change on the wrong branch.🤔

Either way, I’ll fix it asap.

.github/workflows/build-rpi-baremetal.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+  schedule:
+    # Build on Mondays at 9am PST every week
+    - cron:  '0 17 * * 1'
+
+jobs:
+  build-pico-sdk:
```
rauhul, line 13: can you update this name

.github/workflows/build-rpi-baremetal.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-22.04
+    container: swiftlang/swift:nightly-main-jammy
```
rauhul, line 15: Can you use a pinned version of swifts, see the other workflows for examples

rauhul (changes requested):
almost good to merge, sorry about the delay.

.github/workflows/build-rpi-baremetal.yml
iCMDdev: Sure!

iCMDdev:
> almost good to merge, sorry about the delay.

No worries, and happy new year! I made the changes you've requested and retested the workflow, it should work fine now.

rauhul:
@iCMDdev theres a couple yml lint issues remaining, otherwise this is ready

iCMDdev:
Thanks for pointing out!

iCMDdev:
Seems fine now, can you run checks again?

iCMDdev:
Thanks! 🎉
If you want me to change or add anything else, let me know.

kubamracek:
@rauhul shall we merge?

rauhul:
Ah yes please, sorry I thought we already merged it!

I think I got lazy making a nice commit message. The GitHub one wasn't great with all the intermediate commit messages.

Do you mind summarizing the overall change when merging?

iCMDdev:
> The GitHub one wasn't great with all the intermediate commit messages.

Yeah, sorry, I took a bit of advantage of the squash-on-merge policy.

> Do you mind summarizing the overall change when merging?

Perhaps something like this would work:

**Add Raspberry Pi 4B & 5 baremetal examples**
Provide new examples demonstrating how to run Swift Embedded on the Raspberry Pi 4B and 5 single-board computers, in a baremetal fashion, without an underlying operating system. The provided examples showcase how to blink the onboard green LED using Swift MMIO.

iCMDdev:
Thanks for merging! 🎉

PR #69, opened 2024-11-06
Add iBeacon example for Pico W

colemancda:
Updates the Pico W example to enable Bluetooth advertisements, advertising as an iBeacon in addition to its name and Bluetooth address.

Also adds some Swift wrappers for the Pico SDK.

Screenshot 2024-11-06 at 1 45 55 AM


iCMDdev:
Nice work! One remark: I think you changed the wrong description in the README - that’s the non-W pico’s link. 

Also, while I’m not a code owner, I think it would be better to move this code to a separate example from the blink one (just like you did with the ESP example) - consider a beginner just wanting to try out Embedded Swift. I think they’d first want to try out a really simple blink example, and then move on to other more complicated ones like yours.

Again, great work!🎉

colemancda:
@iCMDdev  I updated the branch and moved example to its own folder.

rauhul:
Sorry about the radio silence on your PRs, I'm trying to figure out the best strategy for them. I'm not confident they are the best fit for this repo due to the sheer volume of code and the maintenance burden. I think a better solution would be to host these examples in a fork you own and we can link to them from the community section of the readme. 

thoughts?

PR #73, opened 2024-11-11
doc: add simulation of ESP32-C6 with Wokwi

georgik:
This PR adds files which are necessary for Wokwi simulation to run in the VS Code.
User needs to build the project, then the user can open diagra.json in VS Code with Wokwi plugin and just hit the Play button. This should will start the simulation.

### DIFF
--- a/esp32-led-blink-sdk/diagram.json
+++ b/esp32-led-blink-sdk/diagram.json
@@ -1,6 +1,6 @@
 {
   "version": 1,
-  "author": "Juraj Michálek",
+  "author": "",
   "editor": "wokwi",
   "parts": [
     {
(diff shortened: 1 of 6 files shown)


esp32-led-blink-sdk/diagram.json
```diff
@@ -0,0 +1,36 @@
+{
+  "version": 1,
+  "author": "Juraj Michálek",
```
rauhul, line 3: love the contribution, but could we exclude individual names? (in both json files)

rauhul (approved):

esp32-led-blink-sdk/diagram.json
georgik: @rauhul  good point. Record updated.

PR #78, opened 2024-12-13
Fix error in displaySize

mkbrwr:
`HAL.swift` and `Board.swift` had different names for display width  and display height.
This change renames vars in Board.swift to have the same names as in `HAL.swift`


rauhul:
@kubamracek can you take this one?

kubamracek:
Yup. I can confirm that the build failure is currently present, and that it gets resolved with this PR.

kubamracek (approved):

kubamracek:
...and as this example is using Mach-O files, we can't exercise the build in CI currently.

PR #82, opened 2024-12-20
Add 2 community examples running on nRF52840 microcontroller

ebariaux:


rauhul:
Can you order these alphabetically in the list, ignoring superfluous words like "swift-for"

ebariaux:
I did it per platform, as that seemed to be the logic applied so far

PR #86, opened 2025-01-03
put downloaded swiftc bins at front of path

rauhul:


rauhul:
@kubamracek any idea why the zephyr build is failing?

PR #88, opened 2025-01-14
Add link to nuttx-riscv-blink sample code folder in README

iCMDdev:
Just noticed that the `nuttx-riscv-blink` row inside the the README's Catalog of Examples was missing a link to its sample code folder, presumably due to a typo. This small PR adds a markdown link to `nuttx-riscv-blink` in the README.

Didn't notice this earlier, otherwise I could've bundled it with #67 (which was merged minutes ago).


iCMDdev:
Perhaps adding an example image / screenshot in the table would also be desirable.

kubamracek (approved):

PR #94, opened 2025-03-12
Migrate some stm32 examples to use toolsets

rauhul:
Updates the stm32-uart-echo and stm32-neopixel examples to use a swiftpm toolset.json to specify various compiler and linker flags. Also updates the Makefiles to leverage swiftpm to link the executable instead of manually linking using `ld` directly.


kubamracek (approved):

xtremekforever:
I'm excited for these new features. I'm starting up a new firmware project and I'm giving them a spin- `--toolset`, `--triple`, and so on!

PR #98, opened 2025-03-23
add Arduino LED and Wi-Fi SDKs

### LINKED ISSUE #42: Support for Arduino libraries?
I'm trying to get [arduino-esp32](https://github.com/espressif/arduino-esp32/) to compile using the [esp32-led-blink-sdk](https://github.com/apple/swift-embedded-examples/tree/main/esp32-led-blink-sdk) example as a template, which would be amazing if it worked.

I'm pretty sure it shouldn't be super difficult given how it's literally just a C++ library, but the only issue I've run into is that `swiftc` doesn't seem to like the preprocessor macros that Arduino uses to handle different architectures etc.

Here's the bit of [arduino's CMakeLists.txt](https://github.com/espressif/arduino-esp32/blob/70786dc5fa51601f525496d0f92a220c917b4ad9/CMakeLists.txt#L317-L323) that defines the compiler flags:
```cmake
target_compile_options(${COMPONENT_TARGET} PUBLIC
    -DARDUINO=10812
    -DARDUINO_${idf_target_for_macro}_DEV
    -DARDUINO_ARCH_ESP32
    -DARDUINO_BOARD="${idf_target_caps}_DEV"
    -DARDUINO_VARIANT="${CONFIG_ARDUINO_VARIANT}"
    -DESP32)
```

This is what the `swiftc` call then looks like:
```bash
/usr/bin/swiftc -j 8 -num-threads 8 -c 
-DESP_PLATFORM -DIDF_VER=\"v5.1.4\" -DMBEDTLS_CONFIG_FILE=\"mbedtls/esp_config.h\" -DSOC_MMU_PAGE_SIZE=CONFIG_MMU_PAGE_SIZE -D_GNU_SOURCE -D_POSIX_READER_WRITER_LOCKS 
-parse-as-library -static -emit-module 
-emit-module-path esp-idf/main/__idf_main.swiftmodule
-module-name __idf_main 
-module-link-name main 
-wmo -color-diagnostics
-target riscv32-none-none-eabi
-Xfrontend -function-sections -enable-experimental-feature Embedded -wmo -parse-as-library -Osize 
-Xcc -march=rv32imc_zicsr_zifencei -Xcc -mabi=ilp32 
-pch-output-dir /tmp
-Xfrontend 
-enable-single-module-llvm-emission 
-Xcc -I... -Xcc -I....
-import-bridging-header main/BridgingHeader.h
-cxx-interoperability-mode=default
-DARDUINO=10812 -DARDUINO_ESP32C3_DEV -DARDUINO_ARCH_ESP32 -DARDUINO_BOARD=\"ESP32C3_DEV\" -DARDUINO_VARIANT=\"esp32c3\" -DESP32 
-output-file-map esp-idf/main/CMakeFiles/__idf_main.dir//output-file-map.json 
-I ... -I ....
```

Which raises the following warnings:
```
warning: conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'IDF_VER="v5.1.4"')
warning: conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'MBEDTLS_CONFIG_FILE="mbedtls/esp_config.h"')
warning: conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'SOC_MMU_PAGE_SIZE=CONFIG_MMU_PAGE_SIZE')
warning: conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'ARDUINO=10812')
warning: conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'ARDUINO_BOARD="ESP32C3_DEV"')
warning: conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'ARDUINO_VARIANT="esp32c3"')
```

Which is then used by a lot of Arduino libraries like [OneWire](https://github.com/PaulStoffregen/OneWire/blob/72249e22ef9092b0750e303d266199965a89e500/OneWire.h#L12-L17) for example ([list](https://github.com/search?q=%2F%5E%23if%20ARDUINO%20%3E%3D%2F%20path%3A*.h&type=code)):
```cpp
#if ARDUINO >= 100
#include <Arduino.h>       // for delayMicroseconds, digitalPinToBitMask, etc
#else
#include "WProgram.h"      // for delayMicroseconds
#include "pins_arduino.h"  // for digitalPinToBitMask, etc
#endif
```

Creating an error like this:
```cmake
<unknown>:0: error: failed to emit precompiled header '/tmp/BridgingHeader-swift_1PSIJ0LN6QHW8-clang_1Y9RGGHIW2TB2.pch' for bridging header 'main/BridgingHeader.h'
2 warnings and 4 errors generated.
components/OneWire/OneWire.h:15:10: error: 'WProgram.h' file not found
 13 | #include <Arduino.h>       // for delayMicroseconds, digitalPinToBitMask, etc
 14 | #else
 15 | #include "WProgram.h"      // for delayMicroseconds
    |          `- error: 'WProgram.h' file not found
 16 | #include "pins_arduino.h"  // for digitalPinToBitMask, etc
 17 | #endif
 ```

So it seems that the issue is `swiftc` only supports boolean preproccessor macros; I've tried putting the attributes in my `BridgingHeader.h` as `#define`s but that doesn't seem to work, and I don't know enough about swift's internals how to know what else to try?

Here's what is basically a MNWE (minimal not working example): https://github.com/maartin0/esp32-arduino-swift-test but I'm happy to test any ideas you might have so you don't have to set up a workspace (Arduino is like 1.5 GB)

### PR DESCRIPTION
**Pull request by maartin0**
Adds two esp32 example projects using the [espressif/arduino-esp32](https://github.com/espressif/arduino-esp32/) library to demonstrate Wi-Fi scanning and a simple LED blink example

I've also mentioned this in the [READMEs](https://github.com/maartin0/swift-embedded-examples/blob/arduino-pr/arduino-esp32-wifi-scan-sdk/README.md#requirements): note that these examples require ESP-IDF **v5.3.2** since the [Arduino releases](https://github.com/espressif/arduino-esp32/releases/tag/3.1.3) are paired with specific IDF versions

Resolves #42

### CODE REVIEW

**Comment by ericlewis** [comment]
Blink works but I’ve been unable to get WiFi to work. I’m using a seeed studio C6, so perhaps there are some differences.

**Comment by maartin0** [comment]
> Blink works but I’ve been unable to get WiFi to work. I’m using a seeed studio C6, so perhaps there are some differences. 

Hmm, what errors are you getting? Is it compiling but just not outputting anything?

**Comment by kubamracek** [comment]
The diff looks good, thanks!

Could I ask for a new GH workflow to be added so that the code is built in CI? (Just building is good enough, no need to try to run it.)

**Comment by maartin0** [comment]
I've just written the actions now but there's a dependency issue:
- Swift 6.0.0+ depends on GLIBC_2.38
- The latest [arduino-esp32 release](https://github.com/espressif/arduino-esp32/releases/tag/3.1.3) depends on IDF v5.3 
- The [espressif/idf:v5.3.2](https://hub.docker.com/layers/espressif/idf/v5.3.2/images/sha256-d85ffe6db6162f16589c4c2221de1d5c40c9d9aa2331099e2fe6f5e33d1b8f09) docker image is based on Ubuntu 22.04
- Ubuntu 22.04 bundles GLIBC_2.35

Some options:
1. Use the latest [arduino-esp32 pre-release](https://github.com/espressif/arduino-esp32/releases/tag/3.2.0-RC2) which is based on IDF v5.4 and update the Arduino version when the stable release comes out
2. Manually build the latest version of GLIBC in the action (although this would cause the action to take a lot longer)
3. Don't use the `espressif/idf` docker image and manually install esp-idf in the action (similar to above)
4. Create a custom docker image with IDF v5.3 and ubuntu 24.04

**Comment by maartin0** [comment]
Not sure what you think is best, but I've done (1) which compiles successfully ([example run](https://github.com/maartin0/swift-embedded-examples/actions/runs/14048704707/job/39334920558))

**Comment by eric-humane** [comment]
> > Blink works but I’ve been unable to get WiFi to work. I’m using a seeed studio C6, so perhaps there are some differences.
> 
> Hmm, what errors are you getting? Is it compiling but just not outputting anything?

I didn't realize you were using the hardware serial 😆 
also fun is that Arduino's String collides with swift's string.

I sorta fixed that with this:
```swift
public typealias ArduinoString = String

extension ArduinoString: CustomStringConvertible {
  public var description: Swift.String {
    return Swift.String(cString: self.c_str())
  }
}

extension ArduinoString: Equatable {
  public static func == (lhs: ArduinoString, rhs: ArduinoString) -> Bool {
    lhs.equals(rhs)
  }
  
  public static func == <RHS: StringProtocol>(lhs: ArduinoString, rhs: RHS) -> Bool {
    lhs.equals(String(Swift.String(rhs)))
  }
  
  public static func == <LHS: StringProtocol>(lhs: LHS, rhs: ArduinoString) -> Bool {
    rhs.equals(String(Swift.String(lhs)))
  }
}

extension ArduinoString: Comparable {
  public static func < (lhs: ArduinoString, rhs: ArduinoString) -> Bool {
    lhs.compareTo(rhs) < 0
  }
  
  public static func < <RHS: StringProtocol>(lhs: ArduinoString, rhs: RHS) -> Bool {
    lhs.compareTo(String(Swift.String(rhs))) < 0
  }
  
  public static func < <LHS: StringProtocol>(lhs: LHS, rhs: ArduinoString) -> Bool {
    lhs < rhs.description
  }
}

extension ArduinoString: Hashable {
  public func hash(into hasher: inout Hasher) {
    hasher.combine(description)
  }
}

extension ArduinoString: ExpressibleByStringLiteral {
  public init(stringLiteral value: Swift.String) {
    let cString = Array(value.utf8CString)
    self = ArduinoString(cString)
  }
}

// MARK: - Operators (+, +=)

extension ArduinoString {
  public static func + (lhs: ArduinoString, rhs: ArduinoString) -> ArduinoString {
    var newString = lhs
    _ = newString.concat(rhs)
    return newString
  }
  
  /// ArduinoString + SomeString (e.g. String, Substring, etc.)
  public static func + <RHS: StringProtocol>(lhs: ArduinoString, rhs: RHS) -> ArduinoString {
    ArduinoString(lhs.description + Swift.String(rhs))
  }
  
  /// SomeString + ArduinoString
  public static func + <LHS: StringProtocol>(lhs: LHS, rhs: ArduinoString) -> ArduinoString {
    ArduinoString(Swift.String(lhs) + rhs.description)
  }
  
  /// ArduinoString += ArduinoString
  public static func += (lhs: inout ArduinoString, rhs: ArduinoString) {
    lhs = lhs + rhs
  }
  
  /// ArduinoString += SomeString
  public static func += <RHS: StringProtocol>(lhs: inout ArduinoString, rhs: RHS) {
    lhs = lhs + rhs
  }
}

extension Swift.String {
  public static func += (lhs: inout Swift.String, rhs: ArduinoString) {
    lhs.append(contentsOf: rhs.description)
  }
}

extension ArduinoString {
  // Access the underlying Swift.String's StringProtocol properties
  public var utf8: Swift.String.UTF8View { Swift.String(cString: self.c_str()).utf8 }
  public var utf16: Swift.String.UTF16View { Swift.String(cString: self.c_str()).utf16 }
  public var unicodeScalars: Swift.String.UnicodeScalarView { Swift.String(cString: self.c_str()).unicodeScalars }
  
  // Common StringProtocol methods
  public func hasPrefix(_ prefix: Swift.String) -> Bool {
    return Swift.String(cString: self.c_str()).hasPrefix(prefix)
  }
  
  public func hasSuffix(_ suffix: Swift.String) -> Bool {
    return Swift.String(cString: self.c_str()).hasSuffix(suffix)
  }
  
  public func contains(_ other: Swift.Character) -> Bool {
    return Swift.String(cString: self.c_str()).contains(other)
  }
  
  // Index-based access
  public subscript(position: Swift.String.Index) -> Character {
    return Swift.String(cString: self.c_str())[position]
  }
  
  public subscript(bounds: Range<Swift.String.Index>) -> Swift.Substring {
    return Swift.String(cString: self.c_str())[bounds]
  }
  
  public var startIndex: Swift.String.Index {
    return Swift.String(cString: self.c_str()).startIndex
  }
  
  public var endIndex: Swift.String.Index {
    return Swift.String(cString: self.c_str()).endIndex
  }
  
  public func index(after i: Swift.String.Index) -> Swift.String.Index {
    return Swift.String(cString: self.c_str()).index(after: i)
  }
  
  public func index(before i: Swift.String.Index) -> Swift.String.Index {
    return Swift.String(cString: self.c_str()).index(before: i)
  }
}
```

it's not quite StringProtocol conformance, but does allow for the example to use the usb monitor.

maartin0:
I had to slightly modify that `CustomStringConvertible` implementation since `String#c_str` isn't available without an annotation in the C++ library but that compiles and works quite nicely ([run](https://github.com/maartin0/swift-embedded-examples/actions/runs/14093959625/job/39477161046))

kubamracek:
Can you elaborate on why we need the ArduinoString extensions at all? I.e. where are those extensions used from? I don't immediately see that in the code.

maartin0:
Yes they're not used for anything in the example; as @eric-humane suggested it brings `ArduinoString` closer to conforming to `StringProtocol` but I guess you could just convert them to `Swift.String` and use the existing functionality.

Do you think it would be neater to not include them to keep the example more minimal? The only one that's actually needed is the `CustomStringConvertible` extension.

ericlewis:
> Yes they're not used for anything in the example; as @eric-humane suggested it brings `ArduinoString` closer to conforming to `StringProtocol` but I guess you could just convert them to `Swift.String` and use the existing functionality.
> 
> Do you think it would be neater to not include them to keep the example more minimal? The only one that's actually needed is the `CustomStringConvertible` extension.

That’s a common misconception about CustomStringConvertible actually. String concat won’t work etc. the other issue is it’s sort of “invisible” and hard to understand why it’s occurring. The better thing to do would be to use SWIFT_NAME, probably. But then you’re still having to convert these to swift strings for swift functions and it’s not really clear when or why you have to do that. 

option: conform to string protocol
optimal option: swift rename the arduino string class 
Most optimal option (?): Swift.String supports most of the arduino api string api (it’s much smaller than the option one) and strip out the arduino string class. 

Either way, it’s a tough problem and really hard to debug. Even writing wrappers you run into this a bit.

maartin0:
> That’s a common misconception about CustomStringConvertible actually. String concat won’t work etc. the other issue is it’s sort of “invisible” and hard to understand why it’s occurring. The better thing to do would be to use SWIFT_NAME, probably. But then you’re still having to convert these to swift strings for swift functions and it’s not really clear when or why you have to do that.

Interesting - I've only used string formatting in the example which works fine without the other extensions. I can see why `WiFi.RSSI(i) + "dBm"` might not work for example, but wouldn't `Swift.String(WiFi.RSSI(i)) + "dBm"`, `"\(WiFi.RSSI(i))dBm"` and `"\(WiFi.RSSI(i))" + "dBm"` still work?

> Edit: I’m also not suggesting adding my code above.

Sorry, should have probably waited for your reply. I'll strip out the extensions then

eric-humane:
> > That’s a common misconception about CustomStringConvertible actually. String concat won’t work etc. the other issue is it’s sort of “invisible” and hard to understand why it’s occurring. The better thing to do would be to use SWIFT_NAME, probably. But then you’re still having to convert these to swift strings for swift functions and it’s not really clear when or why you have to do that.
> 
> Interesting - I've only used string formatting in the example which works fine without the other extensions. I can see why `WiFi.RSSI(i) + "dBm"` might not work for example, but wouldn't `Swift.String(WiFi.RSSI(i)) + "dBm"`, `"\(WiFi.RSSI(i))dBm"` and `"\(WiFi.RSSI(i))" + "dBm"` still work?
> 
> > Edit: I’m also not suggesting adding my code above.
> 
> Sorry, should have probably waited for your reply. I'll strip out the extensions then

it would, it just sucks to have to do that :P

maartin0:
I just realised there was a full [arduino-esp32 release](https://github.com/espressif/arduino-esp32/releases/tag/3.2.0) 2 weeks ago based on ESP-IDF v5.4 so I've rebased and updated it to use that version

maartin0:
Hi, just wondering if there were any updates on this?

rauhul:
Do you mind rebasing this so we can more easily review it?

maartin0:
Hi @rauhul, I've renamed and stripped the other examples from my fork ([maartin0/arduino-swift-esp32-examples](https://github.com/maartin0/arduino-swift-esp32-examples)) so it's more or less a standalone repository now. Not sure if you think it's useful to link in the community examples?

PR #106, opened 2025-04-18
pico-blink: build.sh: removes quotes to make it build

chbeer:
build.sh of pico-build didn't build on my Mac (Sequoia) because of the quotes around the $SWIFT_BUILD_FLAGS.

From my understanding, and Swifts understanding, they seem wrong.

### DIFF
--- a/pico-blink/build.sh
+++ b/pico-blink/build.sh
@@ -29,13 +29,13 @@ PYTHON_EXEC=${PYTHON_EXEC:-$(xcrun -f python3)}
 MACHO2UF2=$TOOLSROOT/macho2uf2.py
 
 # Build with Swift package manager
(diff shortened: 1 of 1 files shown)


kubamracek (approved):

kubamracek:
I think I have actually noticed this too, and forgot to fix it... Thanks @chbeer!

rauhul:
I think the right solution is to change `CLANG_FLAGS` and friends into bash arrays not flat strings

rauhul:
@chbeer thank you for this PR! Instead of continuing with this build.sh we can leverage some advancements in SwiftPM to make the build more simple. I've pushed a PR with this new approach in #110.

PR #108, opened 2025-05-01
Use Swiftly to install Swift

rauhul:
Updates the various GitHub workflows to use a common "Install Swift" action using `swiftly` and a `.swift-version` file at the repo root.


kubamracek:
LGTM!

PR #110, opened 2025-05-02
Replace pico-blink build.sh

rauhul:
Removes the legacy build.sh script used to build the pico-blink firmware and replaces it with a makefile leveraging SwiftPM + a new pico.json toolset file. This means we no longer have to handle build flags in the build script directly and SwiftPM can handle invoking the linker.

Supersedes  #106


kubamracek (approved):

kubamracek:
LGTM

PR #137, opened 2025-05-18
Lock Zephyr version to v4.1.0 for nrx-blink-sdk example

xtremekforever:
Using the `main` revision of Zephyr can have unintended side-effects and changes, so it's best to use a stable/tagged version instead.

Also updated the README.md to include the correct `-DBOARD=nrf52840dk/nrf52840` definition for the cmake command, since the previous `-DBOARD=nrf52840dk_nrf52840` is now invalid in the latest version of Zephyr.

@rauhul

### DIFF
--- a/.github/workflows/build-zephyr.yml
+++ b/.github/workflows/build-zephyr.yml
@@ -3,6 +3,8 @@ name: Zephyr
 on:
   push:
     branches: ["main"]
+  pull_request:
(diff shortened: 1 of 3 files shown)


rauhul:
Looks like the build is still failing, :( https://github.com/apple/swift-embedded-examples/actions/runs/15091118649/job/42420159298?pr=137

PR #153, opened 2025-06-09
Improve Integrate with Zephyr doc

xtremekforever:
I re-read the doc and realized some things weren't fully accurate and that some more could be added to help others working with Zephyr and Swift.

 - Fixed the target architecture compatibility to list the correct target triples against Cortex M- cores and ARM architectures.
 - Fixed flags that need to be added to `CMakeLists.txt`, since what was there was not quite accurate.
 - Added section on customizing the linker to strip out unused sections (such as `.swift_modhash`) and link extra libraries such as `swiftUnicodeDataTables`.

Please have a look @ebariaux if you don't mind and see if you spot any issues!


Sources/EmbeddedSwift/Documentation.docc/SDKSupport/IntegrateWithZephyr.md
```diff
@@ -335,3 +350,76 @@ ninja: no work to do.
...
+    swiftUnicodeDataTables
+    -Wl,--no-whole-archive
+    )
+```
+
+Extra code is required to find the right paths where the `swiftUnicodeDataTables.a` file is located, depending on how Swift is installed.
+
+When this is built, depending on the target architecture, warnings may then be printed about 32-bit enums, like this:
+
+```console
```
rauhul, line 421: does this imply you should be using -fshort-enums?

xtremekforever: Well even if you use -fshort-enums in your main Xcc flags, these warnings get printed as soon as you pull in swiftUnicodeDataTables. Could be that somehow the tables are compiled different (like without -fshort-enums) or some other reason.

rauhul: Oh interesting. Regardless, this is a great improvement and we should look into the enum thing independently.

rauhul: Do you mind rebasing and ill run ci?

xtremekforever: Do you think we should open a ticket on swiftlang/swift for this enum mismatch for armv7em? Or just let it be...

Yes I'll rebase in a bit

rauhul: IMO we should just build unicode support as part of the SwiftPM/CMake build graph and avoid prebuilts

PR #157, opened 2025-06-21
Convert stm32-blink to SwiftPM

rauhul:
Replaces bespoke shell scripts with a SwiftPM based build for the stm32-blink example. Adds CI for MachO based examples.

Future: enable ELF variants of all MachO STM32 examples.


rauhul:
> > Adds CI for MachO based examples.
> 
> Very nice. Does that mean this now actually works and we can have Mac hosts in GH actions CI?

Conceptually but I can't get the macOS python install to work

PR #160, opened 2025-06-25
Add a note about unintended _main symbol generation.

andre-richter:


Sources/EmbeddedSwift/Documentation.docc/UsingEmbeddedSwift/Basics.md
```diff
@@ -52,6 +52,8 @@ $ swiftc -target armv7-none-none-eabi -enable-experimental-feature Embedded -wmo
   input1.swift input2.swift ... -c -o output.o
 ```
 
+> Note: Passing only a single `.swift` file to swiftc results in the generation of a `_main` symbol. During linking, this symbol could potentially clash with other objects that also define one. To avoid this, compile multiple Swift files together or use the `-parse-as-library` flag.
```
kubamracek, line 55: Shall we change `_main` to `main`? IIRC, the underscore is only present in Mach-O object files and it's probably more common to use Embedded Swift with ELF targets.

andre-richter: @kubamracek  You're right regarding the `_` not being present for ELF. How about the following to cover both cases?
> results in the generation of a `_main` (Mach-O targets) or `main` (ELF) symbol

@rauhul You mean that it shouldn't make a difference if one or two input files are used (regarding generation of `main`)? It does make a difference in my tests:

```shell
$ swiftc -enable-experimental-feature Embedded -wmo -c -o output.o first.swift && nm output.o | rg main
0000000000000000 T _main

$ swiftc -enable-experimental-feature Embedded -wmo -c -o output.o first.swift second.swift && nm output.o | rg main

```

If that's not what you meant, just let me know which part of the sentence you want me to fix.

rauhul: well im just also not sure about the precise behavior so it would be nice to have @artemcm or someone else more familiar with the driver make sure the documentation is correct

artemcm: The documentation looks right to me, the compiler (the driver actually does not distinguish this) does treat single-input compilation as "script" mode unless `-parse-as-library` is specified or more input source files are added.

andre-richter: Thanks @artemcm 🙏

PR #183, opened 2025-12-23
SVD2LLDB Guided Example

iCMDdev:
This draft PR will add an example on how to use LLDB with Swift Embedded (including [SVD2LLDB](https://forums.swift.org/t/embedded-swift-improvements-coming-in-swift-6-3/)). I decided to create this draft PR and ask for your thoughts & feedback.

Would love to hear what you think, especially regarding:
- How this should be organized (Should it be embedded in one of the current examples? Or, perhaps, a new guide?)
- What board should be used?
- What should the example code do? For example, debugging a timer control register ([like in the SVD2LLDB docs](https://swiftpackageindex.com/apple/swift-mmio/0.1.1/documentation/svd2lldb))


iCMDdev:
I have written the guide and it should be almost ready. The guide was designed for Raspberry Pi Pico (RP2040/RP2350) boards (via conditional compilation using Swift Package traits).

The tutorial guides readers through using LLDB to identify and fix some bugs in a sample I2C embedded app written with Swift MMIO. It is a simple example in which an I2C device, represented by the Pico's `i2c0` interface, is connected to a hypothetical, special I2C peripheral memory, emulated by the `i2c1` interface.

heckj:
:wave: @iCMDdev - I read through the article/tutorial, and would like to provide feedback, but I'll preface this with a very definite I AM NOT AN EMBEDDED DEVELOPER caveat. There may be some baseline expected knowledge that I'm stumbling over.

I made notes _as_ I read through the article, trying to bundle them under section headings. Quite a lot was fairly direct to follow, but there were some places where I was really stumped, unclear on what or how I should do something, or what I should know. In several places, the order of the information presented it later in the section, so I ended up having to read forward, backward, and forward again to get a sense of what I could/should do in each section.

As a general rule, I'd highly suggest removing any of the phrases "simply" or "just" - I've seen that used to make it sound easier or less imposing, but without knowing a lot of this detail already, I felt really silly and awkward.

I love the overall run through the content - using LLDB (through OpenOCD I guess), but when you were presenting the bugs and what was happening, my lack of embedded knowledge and expectations may be very out of whack. I didn't understand many of the intuitive leaps that you made in the writing, although once you made them I could back track and find evidence of bits of it in the presented examples.

- prerequisites are jargon heavy - talk about both GDB and LLDB, and something about an OpenOCD thing as an alternative path - but how (or why) is unclear at this point.
- When you mentioned SVD2LLDB, it was later in the section but kind of needed to be introduced up front from what I read. The link for the installation is points to its docs, but those docs show how you load and use SVD2LLDB, not how to install it - so I was at a loss on what I'd need to do there, and it seems a critical part of what we'll touch on.
- "initial board setup"
  - At end of the section I'd have no idea how to connect up the SWD pins as the instructions ask, and there's no labels in the diagram above it - I'm not sure if the diagram was meant to help there or not. Maybe my lack of basic experience here.
- "uploading the firmware"
  - I found the last sentence almost contradictory - talking about both GDB and LLDB, and I wasn't sure where it was going
> The current example firmware is saved in RAM rather than Flash, so if you unplug your board, you will need to reflash it.
	- (If it didn't go into flash, why would I need to reflash it?) I'm assuming this is an embedded jargon thing that I'm not quite getting - but I get the idea, the thing you uploaded is ephemeral.
- "Attaching LLDB"
	- opens with a warning - but haven't a clue what version is late enough that I need to update. Can any more details be made so I can figure out, other than "it don't work!"
	- I'm guessing from the article about openOCD that it's a bit of code that you can run somewhere, but I'm not sure where that is - is this another RaspberryPi that's set up and connected somehow and forwarding port details? Is that running on my mac? On a Pi? Somewhere else?
	- I'm only barely effective with LLDB, so how to use "thread sel" command, other than the example explicitly provided, isn't meaningful to me. Is there an implication to seeing assembly in one thread, what appears to be C/swift code in another?
	- The instructions are jumbled a bit at the very end of this section, and I'm not sure how they relate to using LLDB earlier. Was the process interupt what let us see the code earlier?
- "Finding bugs with LLDB"
	- implied use `frame sel N` to select the call stack you want to inspect
- "Setting a breakpoint..."
	- just coming clear that reflashing the code onto the board makes it immediately run. Nice detail that you can set a breakpoint BEFORE this happens
	- I have no idea how to identify some of the code as Bootloader code vs. anything else
- "Investigating the second bug"
	- talks to using Control-C to to interupt the code - I'm guessing we're working with at least two terminals, and I'm not clear on which one and where I'd want to invoke that. I'd have to experiment to learn
	- the mismatched registers I get as a bug, but honestly I wouldn't have a clue on what's 'right' and how to resolve it as the tutorial asks me to do.
- "Finding and fixing the last bug"
	- unclear why I need to remember the reset value is 0x65 - and I had no idea as a non-embedded developer.
	- It's later asserted that the i2c1 register has this value, but I'm not seeing it in the examples and wouldn't know how to look right off the bat. I'm not sure if I was supposed to get that from the decode of the complex register or not.

iCMDdev:
Hi @heckj, thanks a lot for the feedback! I would also like to preface this by saying I am not an expert embedded developer either, but rather a student still learning. And this is probably my first time writing documentation this big for a project of this nature, so your advice is greatly appreciated!

First of all, I'd like to note that I added a paragraph withmore information about the setup at the beginning of this guided example, which may make this project easier to digest before reading the prerequisites. I am not sure how I should modify the prerequisites - was striking for a match between keeping things short (explaining in more detail later).

Regarding this:

> When you mentioned SVD2LLDB, it was later in the section but kind of needed to be introduced up front from what I read. The link for the installation is points to its docs, but those docs show how you load and use SVD2LLDB, not how to install it - so I was at a loss on what I'd need to do there, and it seems a critical part of what we'll touch on.

[I changed the link to the correct build instructions.](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L21)

Regarding your other notes, I made some other changes based on your descriptions, so please let me know what you think! I described them below.

#### 1. Removing "simply" / "just"
> As a general rule, I'd highly suggest removing any of the phrases "simply" or "just" - I've seen that used to make it sound easier or less imposing, but without knowing a lot of this detail already, I felt really silly and awkward.

That is definitely not what I was going for, I made a few changes, let me know if this sounds better now.

#### 2. Initial board setup - SWD connection
> - "initial board setup"
>     - At end of the section I'd have no idea how to connect up the SWD pins as the instructions ask, and there's no labels in the diagram above it - I'm not sure if the diagram was meant to help there or not. Maybe my lack of basic experience here.

On the SWD debug probe side, this usually depends on the board and the users' hardware debugger. I tried to clarify this by saying "After you made the connections above, you will need to connect a debugger to the board's SWD pins. Refer to your hardware debugger's documentation for more info". The board's 3 debug pins that need to be connected are those marked as "DEBUG".

#### 3. Uploading the firmware

>> The current example firmware is saved in RAM rather than Flash, so if you unplug your board, you will need to reflash it.
> - (If it didn't go into flash, why would I need to reflash it?) I'm assuming this is an embedded jargon thing that I'm not quite getting - but I get the idea, the thing you uploaded is ephemeral.

Yeah, the idea is indeed that the code is ephemeral the way it's saved currently (RAM memory is erased when the board's power supply is cut), though we could modify the linker scripts to save the code in flash memory (which is persistent). [I changed the wording a bit.](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L123)

#### 4. Attaching LLDB
>    - opens with a warning - but haven't a clue what version is late enough that I need to update. Can any more details be made so I can figure out, other than "it don't work!"

Sorry, yes, this is meant to be updated. This depends on [an OpenOCD patch that I submitted](https://review.openocd.org/c/openocd/+/9461) and is yet to be merged.

>    - I'm guessing from the article about openOCD that it's a bit of code that you can run somewhere, but I'm not sure where that is - is this another RaspberryPi that's set up and connected somehow and forwarding port details? Is that running on my mac? On a Pi? Somewhere else?

Yes, OpenOCD is the program that connects your hardware debugger (an extra piece of hardware connected to your board, like a dongle) to your software debugger, i.e. LLDB. It provides interface that LLDB can use, that is, the GDB remote debug port. [I also made some changes to clarify this.](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L129-L136)

>    - I'm only barely effective with LLDB, so how to use "thread sel" command, other than the example explicitly provided, isn't meaningful to me. Is there an implication to seeing assembly in one thread, what appears to be C/swift code in another?

I added a few words about Swift vs assembly [here](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L192) and [here](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L215). Let me know if you want me to clarify this further - the idea is that one thread is unused (running the chip's own ROM code), while the other one is running our Swift code (for which we have debug symbols + the source code itself).

> The instructions are jumbled a bit at the very end of this section, and I'm not sure how they relate to using LLDB earlier. Was the process interupt what let us see the code earlier

I'm not exactly sure which instructions you're referring to, but yes, when the code execution is interrupted, we can see the code line(s) where the program was stopped - some commands such as `frame sel` / `frame info` / `thread info` display them. You can also use the `source list` command.

#### 5. Setting a breakpoint on our app's entrypoint

>    - I have no idea how to identify some of the code as Bootloader code vs. anything else

[I did describe this previously](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L217) but I am not sure how to specify this without repeating myself; I would love to hear your suggestions.

#### 6. Invesigating the second bug
>    - talks to using Control-C to to interupt the code - I'm guessing we're working with at least two terminals, and I'm not clear on which one and where I'd want to invoke that. I'd have to experiment to learn

This is the terminal in which you're running LLDB. I made some changes to clarify the setup [here](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L136) and [here](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L169). There is only one other terminal, running OpenOCD.

>    - the mismatched registers I get as a bug, but honestly I wouldn't have a clue on what's 'right' and how to resolve it as the tutorial asks me to do.

To be clear, the problem is that `i2c0` is sending data to one address, while the memory device is configured to listen on a different one. As long as these addresses are equal, everything works.

#### 7. Finding and fixing the last bug
>    - unclear why I need to remember the reset value is 0x65 - and I had no idea as a non-embedded developer.

This was meant as "remember for later". [I made some changes, tried to clarify this](https://github.com/iCMDdev/swift-embedded-examples/blob/207e9f99c7e4d60282d96addfb5e91d9692d971a/Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md?plain=1#L562).

>   - It's later asserted that the i2c1 register has this value, but I'm not seeing it in the examples and wouldn't know how to look right off the bat. I'm not sure if I was supposed to get that from the decode of the complex register or not.

It is visible in the decode output (see the second line below):

```
(lldb) svd decode i2c1.IC_CON --read --visual
I2C1.IC_CON: 0x0000_0065 <-- this is the address

                               ╭╴SPEED
                             ╭╴IC_10BITADDR_MASTER
                           ╭╴IC_SLAVE_DISABLE
                         ╭╴TX_EMPTY_CTRL
                       ╭╴STOP_DET_IF_MASTER_ACTIVE
                       ┴ ┴ ┴ ┴ ┴─
0b00000000000000000000000001100101
                        ┬ ┬ ┬ ┬  ┬
                        ╰╴RX_FIFO_FULL_HLD_CTRL
                          ╰╴STOP_DET_IFADDRESSED
                            ╰╴IC_RESTART_EN
                              ╰╴IC_10BITADDR_SLAVE
                                 ╰╴MASTER_MODE

[10:10] STOP_DET_IF_MASTER_ACTIVE 0x0
[9:9]   RX_FIFO_FULL_HLD_CTRL     0x0 (DISABLED)
[8:8]   TX_EMPTY_CTRL             0x0 (DISABLED)
[7:7]   STOP_DET_IFADDRESSED      0x0 (DISABLED)
[6:6]   IC_SLAVE_DISABLE          0x1 (SLAVE_DISABLED)
[5:5]   IC_RESTART_EN             0x1 (ENABLED)
[4:4]   IC_10BITADDR_MASTER       0x0 (ADDR_7BITS)
[3:3]   IC_10BITADDR_SLAVE        0x0 (ADDR_7BITS)
[2:1]   SPEED                     0x2 (FAST)
[0:0]   MASTER_MODE               0x1 (ENABLED)
(lldb)
````

**Comment by iCMDdev** [comment]
The OpenOCD patch was merged yesterday in the following commit: https://github.com/openocd-org/openocd/commit/22e1e1b33e844c343eb3f24d24f630e2471d2b27.

File: Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md [hunk 1]
```diff
@@ -0,0 +1,744 @@
...
+## Prerequisites
+- Install **Swift with Embedded support**, **LLDB**, and **SVD2LLDB** (instructions below)
+- A Raspberry Pi **RP2350 / RP2040**-based board with access to SWD (debug) pins, and an LED. On-board LEDs may be used as long as they are connected directly to GPIO; otherwise, you may use an external LED.
+- A **SWD / JTAG hardware debugger**, allowing GDB remote debug protocol connections via software such as *OpenOCD*.
+
+> Note: If you have another RP2350 / RP2040 board or a Raspberry Pi Debug Probe, you can use it for debugging, via OpenOCD.
+
+
+### Installing Swift, LLDB and SVD2LLDB
+> Note: Embedded Swift is experimental. Public releases of Swift do not support Embedded Swift, yet. See <doc:InstallEmbeddedSwift> for details.
```
heckj: For this guide, I'd suggest minimizing talking about how Embedded Swift is experimental -  anyone popping in here today already has a sense of that, and the key detail in this note is that public releases of Swift don't include Embedded Swift, so you need to install it directly through an open source toolchain. I don't think you need both the call out and the first paragraph, as they're redundant.

The bonus of stripping this note also gives this article better longevity - you won't need to explicitly strip this later when it advances.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+To install Swift for embedded development, follow the instructions in <doc:InstallEmbeddedSwift>, which guides you through using swiftly to install the latest development snapshot with Embedded Swift support. The toolchain includes the LLDB Debugger, so you don't need to install it separately.
+
+To install SVD2LLDB, which is an LLDB plugin that enhances firmware debugging by providing semantic access to hardware registers, follow the instructions [in the swift-mmio docs](https://swiftpackageindex.com/apple/swift-mmio/0.1.1/documentation/svd2lldb/buildingsvd2lldb).
```
heckj:
```suggestion
To install SVD2LLDB, which is an LLDB plugin that enhances firmware debugging by providing semantic access to hardware registers, follow the instructions [in the swift-mmio docs](https://swiftpackageindex.com/apple/swift-mmio/documentation/svd2lldb/buildingsvd2lldb).
```

^ If you strip out the version from the documentation URL, it'll redirect to the latest version.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+Our special memory increments each value that is saved to it.
+
+## Initial board setup
+For this tutorial you will have to connect the `i2c0` and `i2c1` interfaces' `SDA` and `SCL` pins. Use of external pull-ups is recommended.
+
+An example Fritzing schematic for the Raspberry Pi Pico / Pico 2 is provided below:
+
+![Raspberry Pi Pico connection Fritzing diagram](lldb_tutorial.png)
+
+If you are using another RP2040 / RP2350 board, or different pins, you may need to [change the I2C pin numbers in the Application.swift source file]().
```
heckj: missing link?

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+After you made the connections above, you will need to connect a debugger to the board's SWD pins. Refer to your hardware debugger's documentation for more info.
```
heckj:
```suggestion
After you made the connections above, connect a debugger to the board's SWD pins. Refer to your hardware debugger's documentation for more info.
```

stripping out the "you will need to" makes this a bit more direct

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+> Note: If you have another Pico or a Raspberry Pi Debug probe, you can use it for debugging. Check out Raspberry Pi's [docs](https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html) and [picoprobe GitHub repo](https://github.com/raspberrypi/debugprobe) in order to set up the debugger.
+
+## Uploading the firmware
+It's time to run & debug the application.
+
+Connect the Raspberry Pi Pico board via a USB cable to your Mac, and make sure it's in the USB Mass Storage firmware upload mode. This is normally done by *holding the BOOTSEL button while plugging in the board*.
```
heckj:
```suggestion
Connect the Raspberry Pi Pico board via a USB cable to your Mac, and make sure it's in the USB Mass Storage firmware upload mode. This is done by *holding the BOOTSEL button while plugging in the board*.
```

Can we get rid of "normally" here? That begs the question of how it's done otherwise, which isn't addressed.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+drwx------@  1 cmd   staff    16K Jan  9  2007 RP2350
+```
+
+Copy the UF2 file to this volume, either by dragging `Application.uf2` to the new mounted device in Finder or via terminal:
+
+```shell
+$ cp .build/armv7em-apple-none-macho/debug/Application.uf2 /Volumes/RP2350
+```
+
+This will make the Pico automatically install the firmware, reboot itself, and run the firmware. The current example firmware is saved in RAM rather than Flash, so if you unplug your board, the RAM will be cleared, and your program will be lost. Thus, you will need to reflash it.
```
heckj:
```suggestion
Copying in the UF2 file causes the Pico to automatically install the firmware, reboot itself, and run the firmware.
The firmware you uploaded is saved in RAM rather than Flash, so if you unplug your board, the RAM is cleared and your program lost.
In that case, you will need to repeat this process to reflash it.
```

Wording suggestions to make this more direct and explicit

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+## Attaching LLDB
+
+In order to use LLDB, we need to connect it to our board. This is done using a hardware SWD debugger and a compatible debug server program, such as OpenOCD.
```
heckj:
```suggestion
In order to use LLDB, connect it to your board. This is done using a hardware SWD debugger and a compatible debug server program, such as OpenOCD.
```

The writing is jumping back and forth a bit between "we"/"our" and "you"/"your". You used we earlier in the article sort of representing this project or the community in suggesting how to set things up & good patterns, so it gets awkward re-using it as a mechanism for "author + reader" where you're instructing. Most of our documentation uses 2nd person, present tense, active voice for a direct style of instruction that's easy to follow - so I suggest using that in this guide too

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Using OpenOCD
+If your hardware SWD debugger is supported, you can use OpenOCD as a bridge between LLDB and your board.
+
+OpenOCD is a program that can connect your hardware debugger, providing an interface (i.e. the GDB remote debug server) that LLDB (the software debugger program running on your machine) can use to connect and control your board.
```
heckj:
```suggestion
OpenOCD is a program that connects to your hardware debugger, providing an interface that LLDB uses to connect and control your board.
```

^ Is the fact that it's a GDB remote port critical to what we're talking about? GDB isn't referenced elsewhere, and while it's a neat detail, I think it may just add confusion to keep it in when you're focused on using LLDB as the primary debugger.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+> Note: OpenOCD had a bug which may have prevented LLDB from connecting to your board. Ensure you are using the latest version available.
+
+For Pico-based debug probes, using Raspberry Pi's OpenOCD port, you may run the following command in one terminal:
```
heckj:
```suggestion
Use two terminals to set up debugging.
For Pico-based debug probes that use Raspberry Pi's OpenOCD port, run the following command in one of the terminals:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+Info : [rp2350.cm1] Cortex-M33 r1p0 processor detected
+Info : [rp2350.cm1] target has 8 breakpoints, 4 watchpoints
+Info : starting gdb server for rp2350.cm0 on 3333
+Info : Listening on port 3333 for gdb connections
+```
+
+You now have a debug server running on your machine. Leave this terminal open, and do not close the program while you use `lldb` in another terminal.
+
+### Connecting LLDB to your debug server
+Run lldb with your `Application` Mach-O and connect to your debug server (here we are connecting to the OpenOCD server's port, `3333`, as seen in the previous section).
```
heckj:
```suggestion
Run lldb with your `Application` Mach-O and connect to your debug server (here you are connecting to the OpenOCD server's port, `3333`, as seen in the previous section).
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   90  	
+   91  	void interrupt(void) {
+-> 92  	  while (1) {}
+   93  	}
+   94  	
+   95  	__attribute((section("__DATA,stack"), aligned(32)))
+Target 0: (Application) stopped.
+```
+
+We are now connected! If you are seeing assembly output rather than Swift / C code, make sure you provided the correct Mach-O path to LLDB. If you did, then it's worth noting that the Raspberry Pi Pico has 2 cores, exposed to LLDB as 2 threads, and our current example runs on one of them only. The `thread sel` command can be used to select the board's core:
```
heckj:
```suggestion
You are now connected!
If you are seeing assembly output rather than Swift / C code, make sure you provided the correct Mach-O path to LLDB.
If you did, then it's worth noting that the Raspberry Pi Pico has 2 cores, exposed to LLDB as 2 threads, and our current example runs on one of them only.
The `thread sel` command can be used to select the board's core:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   90  	
+   91  	void interrupt(void) {
+-> 92  	  while (1) {}
+   93  	}
+   94  	
+   95  	__attribute((section("__DATA,stack"), aligned(32)))
+(lldb)  
+```
+
+The core that isn’t used by our code is currently idling in the chip's own bootloader (ROM) code. We do not have symbols for the board's ROM, so LLDB can only display the raw assembly instructions.
```
heckj:
```suggestion
The core that isn’t used by our code is currently idling in the chip's own bootloader (ROM) code.
LLDB doesn't have symbols for the board's ROM, so it can only display the raw assembly instructions.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+Otherwise, if you're seeing assembly code in both cases, you might be looking at the board's bootloader (i.e. ROM) code in both cases, and you may not have reached the Swift code yet. A simple way to check this is by looking at the current instruction's address - on RP2040/RP2350, the Boot ROM is memory-mapped starting at `0x00000000`, so any address in the low `0x0000xxxx` range typically lands inside the bootloader. Type `continue` and check that your board isn't waiting for firmware upload (check that it doesn't show up as a mounted volume in `/Volumes`). Then, type `process-interrupt` or hit Control + C (`^C`) to interrupt the program.
```
heckj:
```suggestion
Otherwise, if you're seeing assembly code in both cases, you might be looking at the board's bootloader (that is, its ROM) code.
You may not have reached the Swift code yet.
A simple way to check this is by looking at the current instruction's address - on RP2040/RP2350, the Boot ROM is memory-mapped starting at `0x00000000`, so any address in the low `0x0000xxxx` range is typically inside the bootloader.
Type `continue` and check that your board isn't waiting for firmware upload (check that it doesn't show up as a mounted volume in `/Volumes`).
Then, type `process-interrupt` or hit Control + C (`^C`) to interrupt the program.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+## Finding bugs with LLDB
+
+It is time to find the bugs. If you take a look at the sample application, you will notice that it is supposed to reach an on-board LED-blinking stage; however, you are likely not seeing any turned on LED.
+
+### Uncovering the first issue
+
+Let's find out where the code stopped by checking the backtrace with `thread backtrace` / `bt`:
```
heckj:
```suggestion
Find out where the code stopped by checking the backtrace with `thread backtrace` / `bt`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+    frame #3: 0x20003b72 Application`Application.configureLedPinSIO(Swift.UInt32) -> () at OnboardLED.swift:21:18
+    frame #4: 0x20000968 Application`Application.enableInterfaces() -> () at Application.swift:80:5
+    frame #5: 0x20000bbc Application`static Application.Application.main() -> () at Application.swift:93:5
+    frame #6: 0x20000d6c Application`static Application.Application.$main() -> () at <compiler-generated>:0
+    frame #7: 0x20000d78 Application`Application_main at Application.swift:0
+    frame #8: 0x200053ee Application`reset at Support.c:87:19
+    frame #9: 0x000002f4
+```
+
+It looks like we are dealing with an assertion failure. We are now in an infinite `interrupt()` loop. Frame `#1` is our assertion failure, and frame `#2` seems to be related to Swift MMIO. Meanwhile, from frame `#3` onwards, function calls are related to our Application code itself.
```
heckj:
```suggestion
This looks like an assertion failure.
The code is now in an infinite `interrupt()` loop.
Frame `#1` is our assertion failure, and frame `#2` seems to be related to Swift MMIO.
Meanwhile, from frame `#3` onwards, function calls are related to our Application code itself.
```

(I'm also tending to make each sentence into it's own line in order to make it easier to see the breaks and sentence length)

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+> `interrupt()` serves as a fault handler, since the relevant vector table entries point to it. The vector table is defined in `Support.c`.
+
+Let's see where the code stopped inside Swift MMIO, in frame `#2`. For this, we can use `frame select <frame_number>`:
```
heckj:
```suggestion
To see where the code stopped inside Swift MMIO, in frame `#2`, use `frame select <frame_number>`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   171 	  ) -> Register<Value> where Index: BinaryInteger {
+   172 	    #if hasFeature(Embedded)
+   173 	    // FIXME: Embedded doesn't have static interpolated strings yet
+-> 174 	    precondition(
+   175 	      0 <= index && index < self.count,
+   176 	      "Index out of bounds")
+   177 	    #else
+```
+
+It appears that we are dealing with an out of bounds issue. Fortunately, Swift MMIO has captured it, resulting in an assertion failure rather than a segmentation fault.
```
heckj:
```suggestion
It appears that the code has an out of bounds issue.
Fortunately, Swift MMIO has captured it, resulting in an assertion failure rather than a segmentation fault.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+Let's see what triggered this issue by looking at the frames related to our Application code:
```
heckj:
```suggestion
Find what triggered this issue by looking at the frames related to our Application code:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   91  	struct Application {
+   92  	  static func main() {
+-> 93  	    enableInterfaces()
+   94  	    let controller = I2CController(I2C0_SCL: I2C0_SCL, I2C0_SDA: I2C0_SDA)
+   95  	    let memory = MemoryI2CDevice(I2C1_SCL: I2C1_SCL, I2C1_SDA: I2C1_SDA)
+   96  	
+(lldb)
+```
+
+Looking at the source code, we can see in frame 5 that we were trying to enable our board's relevant interfaces, such as I2C. Frame 3 and 4 suggest that the incorrect out-of-bounds access takes place when the LED's pin is configured. Specifically, `pads_bank0.gpio[pin]` fails, likely due to an incorrect pin number.
```
heckj:
```suggestion
Looking at the source code, you can see in frame 5 that we were trying to enable our board's relevant interfaces, such as I2C. Frame 3 and 4 suggest that the incorrect out-of-bounds access takes place when the LED's pin is configured. Specifically, `pads_bank0.gpio[pin]` fails, likely due to an incorrect pin number.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+Searching our codebase for `LED_PIN`, we can see that it is initialized with `100`:
```
heckj:
```suggestion
Searching the codebase for `LED_PIN`, you can see that it is initialized with `100`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+```swift
+// Board LED
+LED_PIN: UInt32 = 100
+```
+
+The Raspberry Pi Pico has only 40 physical pins, including 26 GPIO; `100` is an odd, incorrect choice for our LED pin number. You will need to change this to your board's correct on-board LED GPIO number, or an external GPIO number of your choice, if you'd rather prefer using an external LED (or if the board doesn't provide a simple on-board LED directly tied to GPIO).
```
heckj:
```suggestion
The Raspberry Pi Pico has only 40 physical pins, including 26 GPIO; `100` is an odd, incorrect choice for the LED pin number.
You will need to change this to your board's correct on-board LED GPIO number, or an external GPIO number of your choice if you'd rather prefer using an external LED (or if the board doesn't provide a simple on-board LED directly tied to GPIO).
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+After making this change, re-build the firmware.
+
+### Setting a breakpoint on our app's entrypoint
+After building the firmware, it's time to re-flash it.
+
+This time, let's try to set up a breakpoint to `static Application.Application.main()`. This helps in certain scenarios, when you don't want to start running code directly, but rather stop at certain instructions.
```
heckj:
```suggestion
This time, try to set up a breakpoint to `static Application.Application.main()`.
This helps in certain scenarios, when you don't want to start running code directly, but rather stop at certain instructions.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+    frame #0: 0x00000000
+->  0x0: movs   r0, r0
+    0x2: and    r0, r0, #0x89
+    0x6: movs   r0, r0
+    0x8: lsls   r1, r7, #0xb
+Target 0: (Application) stopped.
+(lldb) continue
+```
+
+Notice that we have stopped in the board's bootloader code (based on the current instruction's address). Type `continue` to let the bootloader run.
```
heckj:
```suggestion
Notice that the application stopped in the board's bootloader code (based on the current instruction's address). Type `continue` to let the bootloader run.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+-> 93  	    enableInterfaces()
+   94  	    let controller = I2CController(I2C0_SCL: I2C0_SCL, I2C0_SDA: I2C0_SDA)
+   95  	    let memory = MemoryI2CDevice(I2C1_SCL: I2C1_SCL, I2C1_SDA: I2C1_SDA)
+   96  	
+Target 0: (Application) stopped.
+(lldb)  
+```
+
+### Investigating the second bug
+Let's turn our attention to the second bug. If we type `continue`, we will notice that our code doesn't crash, but the LED isn't on, so something is still not right. We suspect our code to be hanging somewhere.
```
heckj:
```suggestion
Turn your attention to the second bug. If you type `continue`, you will notice that our code doesn't crash, but the LED isn't on, so something is still not right. The code is likely hanging somewhere.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   41  	  while true {
+-> 42  	    if cond() { return }
+   43  	    nop()
+   44  	  }
+   45  	}
+```
+
+> Note: This time, when interrupting the program, you may end up stopping at a different place, since the board is continuously looping over some instructions. However, from the third frame 3 onwards, everything should look similar.
+
+It seems that we have stopped in an I2C-related code. Looking at frame 3, we can see that we are waiting for a condition inside `MemoryI2CDevice.swift`. And, by looking at frame `#4`, we can see that our `i2c1` memory peripheral is trying to receive bytes to memory (`Application.MemoryI2CDevice.receiveBytesToMemory()`).
```
heckj:
```suggestion
It seems that the debugger stopped in an I2C-related code.
Looking at frame 3, you can see that it is waiting for a condition inside `MemoryI2CDevice.swift`.
And, by looking at frame `#4`, you can see that the `i2c1` memory peripheral is trying to receive bytes to memory (`Application.MemoryI2CDevice.receiveBytesToMemory()`).
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+We can conclude that somewhere, the `I2C` transmission (between the host `i2c0` and our emulated memory device `i2c1`) hangs. Now it's a good time to try SVD2LLDB's features. Provided that you installed it, load the `SVD` file in the `swift-embedded-examples` repository:
```
heckj:
```suggestion
You can conclude that somewhere, the `I2C` transmission (between the host `i2c0` and our emulated memory device `i2c1`) hangs.
Now it's a good time to try SVD2LLDB's features.
Provided that you installed it, load the `SVD` file in the `swift-embedded-examples` repository:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+```shell
+(lldb) svd load path/to/swift-embedded-examples/Tools/SVDs/rp235x.patched.svd // or rp2040 equivalent
+Loaded SVD file: “rp235x.patched.svd”.
+(lldb)
+```
+
+> Note: An SVD file (System View Description) is an XML description of a microcontroller’s peripherals and registers. Debugging tools like SVD2LLDB and Swift MMIO use them to show named register fields and memory-mapped I/O in a human-friendly way.
+
+Let's inspect the configured I2C addresses by looking at the relevant registers (the one to which `i2c0` writes, and the one on which `i2c1` is listening):
```
heckj:
```suggestion
Inspect the configured I2C addresses by looking at the relevant registers (the one to which `i2c0` writes, and the one on which `i2c1` is listening):
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+  Description: I2C Slave Address Register
+  Address:     0x0000_0000_4009_8008
+  Bit Width:   32
+  Reset Value: 0x0000_0055
+  Reset Mask:  0xffff_ffff
+  Fields:      [IC_SAR]
+(lldb) 
+```
+
+We can use the `svd read` command to read these registers:
```
heckj:
```suggestion
Use the `svd read` command to read these registers:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+RP2350:
+  I2C0:
+    IC_TAR: 0x0000_0043
+(lldb) svd read i2c1.ic_sar
+RP2350:
+  I2C1:
+    IC_SAR: 0x0000_0055
+```
+
+We can clearly see a difference here:  `i2c0`'s target address is `0x43`, while `i2c1` is holding its reset value, `0x55`. Now it's a good idea to check what value `i2c1` is supposed to be listening on, by looking in `MemoryI2CDevice.swift`:
```
heckj:
```suggestion
You can clearly see a difference here:  `i2c0`'s target address is `0x43`, while `i2c1` is holding its reset value, `0x55`.
Now it's a good idea to check what value `i2c1` is supposed to be listening on, by looking in `MemoryI2CDevice.swift`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+```swift
+    // Set peripheral address
+    i2c1.ic_sar.write { w in
+        w.storage = 0x42
+    }
+```
+
+It seems that `i2c1` is supposed to be listening on `0x42`. Therefore, part of the issue may be caused by the mismatch between the `i2c0`'s target address `0x43`, and `i2c1`'s listening address, which is supposed to be `0x42`. However, as we previously mentioned, `i2c1` is currently set to `0x55`, so another bug might be present.
```
heckj:
```suggestion
It seems that `i2c1` is supposed to be listening on `0x42`. Therefore, part of the issue may be caused by the mismatch between the `i2c0`'s target address `0x43`, and `i2c1`'s listening address, which is supposed to be `0x42`. However, as previously mentioned, `i2c1` is currently set to `0x55`, so another bug might be present.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+(lldb) svd read i2c0.ic_tar                                                                                                                                                                    RP2350:
+  I2C0:
+    IC_TAR: 0x0000_0042
+(lldb) svd read i2c1.ic_sar
+RP2350:
+  I2C1:
+    IC_SAR: 0x0000_0055
+```
+
+As you can recall, we already suspected that we fixed only one part of this issue. It's now time to check out other relevant I2C MMIO registers.
```
heckj:
```suggestion
Earlier the article mentioned there may be more than one issue.
It's now time to check out other relevant I2C MMIO registers.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+At this stage, we recommend opening your board's datasheet ([RP2040 datasheet - section 4.3.17](https://pip-assets.raspberrypi.com/categories/814-rp2040/documents/RP-008371-DS-1-rp2040-datasheet.pdf) / [RP2350 datasheet - section 12.2.17](https://pip-assets.raspberrypi.com/categories/1214-rp2350/documents/RP-008373-DS-2-rp2350-datasheet.pdf)).
```
heckj:
```suggestion
At this stage, look at your board's datasheet ([RP2040 datasheet - section 4.3.17](https://pip-assets.raspberrypi.com/categories/814-rp2040/documents/RP-008371-DS-1-rp2040-datasheet.pdf) / [RP2350 datasheet - section 12.2.17](https://pip-assets.raspberrypi.com/categories/1214-rp2350/documents/RP-008373-DS-2-rp2350-datasheet.pdf)).
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+Let's turn our heads to `IC_CON`, the I2C Control Register. This is main register responsible for our I2C interface configuration. It is also the first register in the datasheet's I2C registers list.
```
heckj:
```suggestion
Find `IC_CON`, the I2C Control Register.
This is main register responsible for our I2C interface configuration.
It is also the first register in the datasheet's I2C registers list.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+  Address:     0x0000_0000_4009_8000
+  Bit Width:   32
+  Reset Value: 0x0000_0065
+  Reset Mask:  0xffff_ffff
+  Fields:      [STOP_DET_IF_MASTER_ACTIVE, RX_FIFO_FULL_HLD_CTRL, TX_EMPTY_CTRL, STOP_DET_IFADDRESSED, IC_SLAVE_DISABLE, IC_RESTART_EN, IC_10BITADDR_MASTER, IC_10BITADDR_SLAVE, SPEED, MASTER_MODE]
+```
+
+Remember that the reset value for these registers is 0x65 - this can help you identify whether the register was modified or is at its default value. This value is also indicated in the datasheet.
+
+Since this is a complex register with multiple fields, it is a great candidate for the `decode` SVD2LLDB command. We can read the register and display its value in a beautiful, visual way:
```
heckj:
```suggestion
Since this is a complex register with multiple fields, it is a great candidate for the `decode` SVD2LLDB command.
With it, you can read the register and display its value in a beautiful, visual way:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+[4:4]   IC_10BITADDR_MASTER       0x0 (ADDR_7BITS)
+[3:3]   IC_10BITADDR_SLAVE        0x0 (ADDR_7BITS)
+[2:1]   SPEED                     0x2 (FAST)
+[0:0]   MASTER_MODE               0x1 (ENABLED)
+(lldb)
+```
+
+> Note: You can also read the register manually using LLDB `memory read` commands, and provide its value as an argument. For more info, run `svd decode --help`.
+
+It seems that the `i2c0` interface is correctly initialized, while the `i2c1` register doesn't seem to be configured at all, containing its reset value. This is likely why the reading operation stalls. In order to find out why this is actually happening, we should take a closer look at `MemoryI2CDevice.swift`, and the function responsible for I2C bus configuration:
```
heckj:
```suggestion
It seems that the `i2c0` interface is correctly initialized, while the `i2c1` register doesn't seem to be configured at all, containing its reset value.
This is likely why the reading operation stalls.
In order to find out why this is actually happening, take a closer look at `MemoryI2CDevice.swift`, and the function responsible for I2C bus configuration:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+        // Set peripheral address
+        i2c1.ic_sar.write { w in
+            w.storage = address
+        }
+    }
+}
+```
+
+At first glance, this code may look reasonable - the I2C bus is enabled, and then configured using `configBus()`. Let's try to set a breakpoint on `configBus` and re-run our app, to see if this write really fails.
```
heckj:
```suggestion
At first glance, this code may look reasonable - the I2C bus is enabled, and then configured using `configBus()`.
Set a breakpoint on `configBus` and re-run our app, to see if this write really fails.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   41  	
+   42  	        // Config as peripheral
+-> 43  	        i2c1.ic_con.write { w in
+   44  	            w.raw.master_mode = 0
+   45  	            w.raw.speed = 1
+   46  	            w.raw.ic_restart_en = 1
+(lldb)
+```
+
+We have now reached the `configBus()` function. Let's check the IC_CON value before writing:
```
heckj:
```suggestion
You have now reached the `configBus()` function. Check the IC_CON value before writing:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+```
+(lldb) svd read i2c1.ic_con
+RP2350:
+  I2C1:
+    IC_CON: 0x0000_0065
+```
+
+And now, if we step-over the register write, and check if the new values were written:
```
heckj:
```suggestion
Step-over the register write and check if the new values were written:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+   53  	        }
+   54  	    }
+Target 0: (Application) stopped.
+(lldb) svd read i2c1.ic_con
+RP2350:
+  I2C1:
+    IC_CON: 0x0000_0065
+```
+
+We can see that nothing changes. Therefore, the issue relies on this register write - somehow, our value isn't written to the register at all. At this point, it's a good idea to check the datasheet again, asking ourselves *"are we writing this wrong?"*.
```
heckj:
```suggestion
You can see that nothing changes. Therefore, the issue relies on this register write - somehow, the value isn't written to the register at all. At this point, it's a good idea to check the datasheet again, asking yourself *"are we writing this wrong?"*.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+
+> IC_CON Register Description, according to the datasheet:
+> This register can be written only when the DW_apb_i2c is disabled, which corresponds to the IC_ENABLE[0] register being set to 0. Writes at other times have no effect. 
+
+This must be our problem, since it matches our symptoms! Therefore, we cannot write to the register, since the `i2c1` interface isn't disabled when we configure it inside `MemoryI2CDevice.swift`:
```
heckj:
```suggestion
This must be the problem, since it matches our symptoms! Therefore, it can't write to the register, since the `i2c1` interface isn't disabled when you configure it inside `MemoryI2CDevice.swift`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,744 @@
...
+        // configBus()
+
+        // ✅ After
+        disableI2C()
+        configBus()
+        enableI2C()
+    }
+```
+
+Therefore, we must first call `disableI2C()` before configuring the bus. The host `i2c0` interface already does this in `I2CController.swift`, so that's why it was correctly configured from the beginning.
```
heckj:
```suggestion
Therefore, you must first call `disableI2C()` before configuring the bus. The host `i2c0` interface already does this in `I2CController.swift`, so that's why it was correctly configured from the beginning.
```

heckj (commented):
The updates for explanations in the guide are great! I ran through the whole thing again this morning, and stumbled a bit with the back and forth between the tenses (jumping between we/our and you/your) - so I've made a (large) set of suggestions for changing most of the wording to use you/your, which matches the style that's common in Swift's documentation.

I may not have hit them all, but hopefully you get a sense of it. The place where I explicitly left "we" alone was in the earlier part of the article where "we" seemed to imply this project/the community and was providing recommendations. 

I also dropped some of the "You can ..." to be more direct in getting to the action of what the guide is asking the reader to do. Read through with these changes, see what you think. I tried to not make any changes that significantly removed your voice from the writing, while making it more direct and consistent.

heckj (commented):
The updates for explanations in the guide are great! I ran through the whole thing again this morning, and stumbled a bit with the back and forth between the tenses (jumping between we/our and you/your) - so I've made a (large) set of suggestions for changing most of the wording to use you/your, which matches the style that's common in Swift's documentation.

I may not have hit them all, but hopefully you get a sense of it. The place where I explicitly left "we" alone was in the earlier part of the article where "we" seemed to imply this project/the community and was providing recommendations. 

I also dropped some of the "You can ..." to be more direct in getting to the action of what the guide is asking the reader to do. Read through with these changes, see what you think. I tried to not make any changes that significantly removed your voice from the writing, while making it more direct and consistent.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
iCMDdev: Makes sense - thanks!

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
iCMDdev: Fair point! Yes, it's not critical.

My original thought was to justify why we're using `(lldb) gdb-remote 3333` later (or what platform should be selected if using `platform connect`), but I now realize that we can avoid it to stay clear of any possible confusions.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
iCMDdev: Makes sense, looks much cleaner this way!

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+Our special memory increments each value that is saved to it.
+
+## Initial board setup
+For this tutorial you will have to connect the `i2c0` and `i2c1` interfaces' `SDA` and `SCL` pins. Use of external pull-ups is recommended.
+
+An example Fritzing schematic for the Raspberry Pi Pico / Pico 2 is provided below:
+
+![Raspberry Pi Pico connection Fritzing diagram](lldb_tutorial.png)
+
+If you are using another RP2040 / RP2350 board, or different pins, you may need to [change the I2C pin numbers in the Application.swift source file]().
```
iCMDdev:
```suggestion
If you are using another RP2040 / RP2350 board, or different pins, you may need to [change the I2C pin numbers in the Application.swift source file](https://github.com/swiftlang/swift-embedded-examples/blob/main/rpi-pico-lldb/start-tutorial/Sources/Application/Application.swift).
```

Note: The link is currently broken, as it points to where the file will be when this PR is merged. It should work after this merging this PR.

rpi-pico-lldb/finished-tutorial/README.md
```diff
@@ -0,0 +1,7 @@
+# Finished tutorial
+This is the bugs-fixed version sample code.
+
+Bugs fixed:
+- In `MemoryI2CDevice.init(I2C1_SCL: UInt32, I2C1_SDA: UInt32, address: UInt32, enableInternalPullUp: Bool)`,  `disableI2C()` must be called before calling `configBus()`.According to the RP2350/RP2040 datasheet, setting the `ic_con` and `ic_sar` registers requires the I2C interface to be disabled beforehand.
+- In `Application.main()`, an incorrect address of the I2C peripheral was used (`controller.configBus(targetPeripheral: 0x43)`). We've changed it to the correct address, `0x42`.
```
iCMDdev:
```suggestion
- In `Application.main()`, an incorrect address of the I2C peripheral was used (`controller.configBus(targetPeripheral: 0x43)`). It was changed to the correct address, `0x42`.
```

rpi-pico-lldb/finished-tutorial/README.md
```diff
@@ -0,0 +1,7 @@
...
+- The on-board LED's pin (`LED_PIN`) was incorrect (originally `100`). We changed it to the correct pin of the Pico's on-board LED (`25`).
```
iCMDdev:
```suggestion
- The on-board LED's pin (`LED_PIN`) was incorrect (originally `100`). This was changed to the correct pin of the Pico's on-board LED (`25`).
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+   95  	    let memory = MemoryI2CDevice(I2C1_SCL: I2C1_SCL, I2C1_SDA: I2C1_SDA)
+   96  	
+Target 0: (Application) stopped.
+(lldb)  
+```
+
+### Investigating the second bug
+Turn your attention to the second bug. If you type `continue`, you will notice that our code doesn't crash, but the LED isn't on, so something is still not right. The code is likely hanging somewhere.
+
+Let's interrupt the code with Control + C (`^C`) or `process interrupt`:
```
iCMDdev:
```suggestion
Interrupt the code with Control + C (`^C`) or `process interrupt`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+-> 174 	    precondition(
+   175 	      0 <= index && index < self.count,
+   176 	      "Index out of bounds")
+   177 	    #else
+```
+
+It appears that the code has an out of bounds issue.
+Fortunately, Swift MMIO has captured it, resulting in an assertion failure rather than a segmentation fault.
+
+Find what triggered this issue by looking at the frames related to our Application code:
```
iCMDdev:
```suggestion
Find what triggered this issue by looking at the frames related to the Application code:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+  I2C1:
+    IC_CON: 0x0000_0065
+```
+
+You can see that nothing changes. Therefore, the issue relies on this register write - somehow, the value isn't written to the register at all. At this point, it's a good idea to check the datasheet again, asking yourself *"are we writing this wrong?"*.
+
+> IC_CON Register Description, according to the datasheet:
+> This register can be written only when the DW_apb_i2c is disabled, which corresponds to the IC_ENABLE[0] register being set to 0. Writes at other times have no effect. 
+
+This must be the problem, since it matches our symptoms! Therefore, it can't write to the register, since the `i2c1` interface isn't disabled when you configure it inside `MemoryI2CDevice.swift`:
```
iCMDdev:
```suggestion
This must be the problem, since it matches the symptoms! Therefore, it can’t write to the register, since the `i2c1` interface isn’t disabled when you configure it inside `MemoryI2CDevice.swift`:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+        // Set peripheral address
+        i2c1.ic_sar.write { w in
+            w.storage = address
+        }
+    }
+}
+```
+
+At first glance, this code may look reasonable - the I2C bus is enabled, and then configured using `configBus()`.
+Set a breakpoint on `configBus` and re-run our app, to see if this write really fails.
```
iCMDdev:
```suggestion
Set a breakpoint on `configBus` and re-run the app, to see if this write really fails.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+    IC_SAR: 0x0000_0055
+```
+
+Earlier the article mentioned there may be more than one issue.
+It's now time to check out other relevant I2C MMIO registers.
+
+At this stage, look at your board's datasheet ([RP2040 datasheet - section 4.3.17](https://pip-assets.raspberrypi.com/categories/814-rp2040/documents/RP-008371-DS-1-rp2040-datasheet.pdf) / [RP2350 datasheet - section 12.2.17](https://pip-assets.raspberrypi.com/categories/1214-rp2350/documents/RP-008373-DS-2-rp2350-datasheet.pdf)).
+
+Find `IC_CON`, the I2C Control Register.
+This is main register responsible for our I2C interface configuration.
```
iCMDdev:
```suggestion
This is main register responsible for the I2C interface configuration.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+   45  	}
+```
+
+> Note: This time, when interrupting the program, you may end up stopping at a different place, since the board is continuously looping over some instructions. However, from the third frame 3 onwards, everything should look similar.
+
+It seems that the debugger stopped in an I2C-related code.
+Looking at frame 3, you can see that it is waiting for a condition inside `MemoryI2CDevice.swift`.
+And, by looking at frame `#4`, you can see that the `i2c1` memory peripheral is trying to receive bytes to memory (`Application.MemoryI2CDevice.receiveBytesToMemory()`).
+
+You can conclude that somewhere, the `I2C` transmission (between the host `i2c0` and our emulated memory device `i2c1`) hangs.
```
iCMDdev:
```suggestion
You can conclude that somewhere, the `I2C` transmission (between the host `i2c0` and the emulated memory device `i2c1`) hangs.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+-> 93  	    enableInterfaces()
+   94  	    let controller = I2CController(I2C0_SCL: I2C0_SCL, I2C0_SDA: I2C0_SDA)
+   95  	    let memory = MemoryI2CDevice(I2C1_SCL: I2C1_SCL, I2C1_SDA: I2C1_SDA)
+   96  	
+Target 0: (Application) stopped.
+(lldb)  
+```
+
+### Investigating the second bug
+Turn your attention to the second bug. If you type `continue`, you will notice that our code doesn't crash, but the LED isn't on, so something is still not right. The code is likely hanging somewhere.
```
iCMDdev:
```suggestion
Turn your attention to the second bug. If you type `continue`, you will notice that your code doesn't crash, but the LED isn't on, so something is still not right. The code is likely hanging somewhere.
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+(lldb) continue
+```
+
+Notice that the application stopped in the board's bootloader code (based on the current instruction's address). Type `continue` to let the bootloader run.
+
+Then, upload the `UF2` firmware to the new mounted volume.
+
+> Note: There are other ways to upload the binary as well. For example, you can also copy the binary directly over SWD, rather than manually uploading it via the mounted volume.
+
+Right after uploading the firmware, you will reach our entrypoint breakpoints:
```
iCMDdev:
```suggestion
Right after uploading the firmware, you will reach your entrypoint breakpoints:
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
iCMDdev: Makes sense. This warning is also present in these other (currently published on swift.org) pages:
- [Introduction to Embedded Swift](https://docs.swift.org/embedded/documentation/embedded/introduction)
- [Getting started with Embedded Swift](https://docs.swift.org/embedded/documentation/embedded/waystogetstarted)
- [Raspberry Pi Pico Blink (Pico SDK) guide](https://docs.swift.org/embedded/documentation/embedded/picoguide)

Should we remove it from the Pico example? I think the first two make sense, since these are pages with general information, but once the reader approaches the precise guided examples, they already know this. 

Moreover, this warning isn't present in the [macOS Guided Example](https://docs.swift.org/embedded/documentation/embedded/macosguide) or the [STM32 one](https://docs.swift.org/embedded/documentation/embedded/stm32baremetalguide).

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+## Attaching LLDB
+
+In order to use LLDB, connect it to your board. This is done using a hardware SWD debugger and a compatible debug server program, such as OpenOCD.
+
+### Using OpenOCD
+If your hardware SWD debugger is supported, you can use OpenOCD as a bridge between LLDB and your board.
+
+OpenOCD is a program that connects to your hardware debugger, providing an interface that LLDB uses to connect and control your board.
+
+> Note: OpenOCD had a bug which may have prevented LLDB from connecting to your board. Ensure you are using the latest version available.
```
iCMDdev:
```suggestion
> Note: OpenOCD had a bug which may have prevented LLDB from connecting to your board, which was fixed in commit [22e1e1b](https://review.openocd.org/q/22e1e1b33e844c343eb3f24d24f630e2471d2b27). Ensure the version you are using includes this patch. The Raspberry Pi fork's afferent commit is [acff23f](https://github.com/raspberrypi/openocd/commit/acff23ffd100479d6481a1155123774b79a990b9).
```

Note to myself: this should be updated to the latest state when this guide is published (that is, include a version number rather than a commit hash, if available). Current state is that the OpenOCD patch was merged in the aforementioned commit, but there's no tagged release containing it yet (maintainers said version 1.0.0 should be tagged soon / asap).

Also, this commit is not yet present in the Raspberry Pi fork (for the Pico / RPi Debug Probe) - linking raspberrypi/openocd#147 for reference.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
heckj: I don't think we need to go out of our way on this topic to normalize it - it'll be more places to switch once it *is* part of a release, my suggestion for this guide just reduces that future workload by 1 ;-)

rauhul: FWIW we should only say this once in the entire documentation catalog. Each article repeating it isn't helpful. 

Additionally, that warning was introduced when embedded would crash the compiler extremely frequently. It's no longer a super helpful warning, as we're well on the way to productizing.

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
```diff
@@ -0,0 +1,773 @@
...
+## Prerequisites
+- Install **Swift with Embedded support**, **LLDB**, and **SVD2LLDB** (instructions below)
+- A Raspberry Pi **RP2350 / RP2040**-based board with access to SWD (debug) pins, and an LED. On-board LEDs may be used as long as they are connected directly to GPIO; otherwise, you may use an external LED.
+- A **SWD / JTAG hardware debugger**, allowing GDB remote debug protocol connections via software such as *OpenOCD*.
+
+> Note: If you have another RP2350 / RP2040 board or a Raspberry Pi Debug Probe, you can use it for debugging, via OpenOCD.
+
+
+### Installing Swift, LLDB and SVD2LLDB
+> Note: Embedded Swift is experimental. Public releases of Swift do not support Embedded Swift, yet. See <doc:InstallEmbeddedSwift> for details.
```
iCMDdev:
```suggestion
```

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
iCMDdev: I removed it from the current guide (can mark this as solved). So, do you want me to keep the others, remove them in another PR or should I rather remove them here?

> Additionally, that warning was introduced when embedded would crash the compiler extremely frequently. It's no longer a super helpful warning, as we're well on the way to productizing.

It does indeed feel very robust and much less experimental nowadays :)

Sources/EmbeddedSwift/Documentation.docc/GuidedExamples/LLDBGuide.md
iCMDdev: The Raspberry Pi fork contains the patch in commit [acff23f](https://github.com/raspberrypi/openocd/commit/acff23ffd100479d6481a1155123774b79a990b9).

It was there since last month, but only noticed this today, sorry!

rauhul:
@heckj is there a good way to host this sample code as a downloadable package on the docc page?

rauhul:
@iCMDdev is it possible to build these start/finish states in ci?

iCMDdev:
Sure, I will write some ci tests ASAP.

iCMDdev:
Hi @rauhul, forgot to tell you that I added CI builds - let me know what you think!

PR #192, opened 2026-02-03
ESP: fix build issues by introducing swift component

Lapshin:
@erhankur , this change fixes the issue with picolibc enabled that we discussed.

The root cause was adding `CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES` into the main target. It makes the include argument order unexpected.
When the inclusion was moved to a separate component, it respects includes from `esp_libc` that should be first.

IMO it's better to have `swift` in managed components. Could you take care of this?

Other observations:

- `BridgingHeader.h` can be generated automatically (include all public headers from required components) 
- LED does not blink on my esp32c6 board. Even with the code before my changes


rauhul:
I would prefer if this change moved the esp examples into an `esp/` directory and the components was placed under there

erhankur:
@Lapshin Now we have one more example. `esp32-uart-echo`

Lapshin:
depends on https://github.com/espressif/idf-extra-components/pull/685

esp32-led-blink-sdk/main/idf_component.yml
```diff
@@ -0,0 +1,2 @@
+dependencies:
+  espressif/swift: "^1.0.0"
```
erhankur: We can run the tests with that PR until its merged
```  
swift:
    git: https://github.com/espressif/idf-extra-components.git
    path: swift
    version: "refs/pull/685/head"
```

erhankur: And I think we can continue to keep the examples under the root (not esp/) since we will not need any common directory. managed_components will be installed inside each examples.

Lapshin: Ok, moved them back

Lapshin: Let's wait until https://github.com/espressif/idf-extra-components/pull/685 merged. I checked examples locally

erhankur:
@Lapshin LGTM Nice cleanup. Tested all examples.

erhankur (approved):

Lapshin:
@rauhul , PTAL

Required component is in its place https://components.espressif.com/components/espressif/idf_swift

erhankur:
@rauhul this is ready for review.

erhankur:
@rauhul Build issue is fixed. Once more PTAL.

rauhul (approved):
Awesome!

rauhul:
Will merge after checks pass after rebasing

PR #197, opened 2026-03-28
Upgrade to a more recent version of Zephyr.

MMcM:
This shows some new incompatibilities.

Specifically, https://github.com/zephyrproject-rtos/zephyr/commit/c4c1d92ceee1357820adc0786e595d06e3eea4d0#diff-d31e6c0e092586b6cc791077b8aaee7c14737ad84e51d4ef943000d062842ece changed the GPIO bit-fields to be function macros in `4.2`, breaking the bridging header.

### DIFF
--- a/.github/workflows/build-zephyr.yml
+++ b/.github/workflows/build-zephyr.yml
@@ -27,14 +27,14 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.11
+          python-version: 3.12
 
       - name: Setup Zephyr project
         uses: zephyrproject-rtos/action-zephyr-setup@v1
         with:
           app-path: ${{ matrix.example }}
           toolchains: arm-zephyr-eabi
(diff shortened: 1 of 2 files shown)


xtremekforever:
> This shows some new incompatibilities.
> 
> Specifically, [zephyrproject-rtos/zephyr@c4c1d92#diff-d31e6c0e092586b6cc791077b8aaee7c14737ad84e51d4ef943000d062842ece](https://github.com/zephyrproject-rtos/zephyr/commit/c4c1d92ceee1357820adc0786e595d06e3eea4d0#diff-d31e6c0e092586b6cc791077b8aaee7c14737ad84e51d4ef943000d062842ece) changed the GPIO bit-fields to be function macros in `4.2`, breaking the bridging header.

I noticed this as well when I tried a newer Zephyr. Now the big question is- what is the best way to deal with these bit field changes? I redefined these macros as an enum on the Swift side, but maybe there's another way.

PR #227, opened 2026-07-14
Fix documentation drift in DocC guides and README

DrPitre:
## Summary

An audit of the DocC catalog (`Sources/EmbeddedSwift/EmbeddedSwift.docc`) and the top-level README against the current state of the example projects turned up documentation that had drifted from the source it describes. Every fix below was verified against the real example source, Makefiles, or `swiftc`/`swift build` output before being applied — not just inferred.

- Fix GPIO pin number in ESP32LedStripGuide (doc said pin 0, code/hardware use pin 8)
- Fix "Modfy" typo in ESP32UartEchoGuide
- Fix three invalid Swift syntax errors in Existentials.md snippets (confirmed with `swiftc -typecheck`)
- Fix wrong ARMv8-M target triple in IntegrateWithZephyr's architecture table (was `aarch64-none-none-elf`, a 64-bit AArch64 triple; should be `armv8m.main-none-none-eabi` for Cortex-M23/M33/M85) and add the missing `hal_nordic` dependency to its sample `west.yml`, matching the real `nrfx-blink-sdk/west.yml`
- Update stale `.build/...` paths across the STM32 and Raspberry Pi baremetal guides to match the actual triple-specific paths their Makefiles produce via `swift build --show-bin-path`
- Fix ~10 stale line-number references and a missing `let` in LLDBGuide's lldb transcripts (the tutorial source was edited after the guide was written) and fix a wrong directory name (`rpi-pico2-lldb` → `rpi-pico-lldb`)
- Genericize a hardcoded, now-stale snapshot date in InstallEmbeddedSwift
- Add the missing "requires a nightly toolchain" caveat to macOSGuide, present in every sibling guide
- Note the full `rpi-pico-blink-sdk` example from IntegrateWithPico's minimal walkthrough
- Add the missing `rpi-pico-lldb` entry to the README example catalog

## Test plan

- [x] Verified GPIO pin against `esp32-led-strip-sdk/main/Main.swift`
- [x] Verified Existentials.md snippets fail/succeed as expected with `swiftc -typecheck`
- [x] Verified target triples against `swift build --triple <t> --show-bin-path` recognition
- [x] Verified `.build` paths against each example's Makefile and a real `swift build --show-bin-path` run
- [x] Verified LLDBGuide line numbers against `rpi-pico-lldb/start-tutorial` and `finished-tutorial` sources
- [x] Confirmed all edited markdown files have balanced code fences


Sources/EmbeddedSwift/EmbeddedSwift.docc/Examples/RPi4bBlinkGuide.md
```diff
@@ -56,14 +56,14 @@ $ cd rpi-4b-blink
 $ make
 ```
 
-This produces `.build/release/Application.bin`, the raw kernel image extracted from the built ELF executable.
+This produces `.build/aarch64-none-none-elf/release/Application.bin`, the raw kernel image extracted from the built ELF executable.
```
rauhul, line 59: `build/release/Application` is correct!

DrPitre: I just re-verified this locally before assuming the fix was wrong. Running the exact command the `Makefile` uses:

```
$ swift build --build-system native --configuration release --triple aarch64-none-none-elf --toolset Tools/Toolsets/rpi-5-elf.json --show-bin-path
.../rpi-4b-blink/.build/aarch64-none-none-elf/release
```

There's no `.build/release` directory produced at all in this package — only `.build/aarch64-none-none-elf/release` (the cross-compiled target build) and `.build/arm64-apple-macosx/release` (host-side build tools/macros). That's what prompted the original fix.

Is there a different build path you're seeing on your end (different SwiftPM version, different toolset, or maybe testing with a plain `swift build` without `--triple`)? Happy to adjust if I'm missing something, but wanted to share what I found before reverting.

DrPitre: image

PR #231, opened 2026-07-15
Add article on porting Embedded Swift to a new platform

DrPitre:
## Summary
- Adds a new DocC article, `PortingToNewPlatforms.md`, documenting the Embedded Swift Platform Abstraction Layer (`EmbeddedPlatform.h`) and the runtime entry points a platform must provide (allocation, console output, randomness, mutexes, exclusivity checking, exit).
- Points readers at the ready-made shims in [`stdlib/public/EmbeddedPlatform`](https://github.com/swiftlang/swift/tree/main/stdlib/public/EmbeddedPlatform) (POSIX, single-threaded, multi-threaded POSIX/Darwin) that they can reuse or copy as a starting point.
- Links the new article from the landing page's "SDK support" section, alongside `IntegratingWithPlatforms` and `Baremetal`.

## Test plan
- [x] Verified every function name, signature, and flag mentioned (`_swift_allocate`, `_swift_mutex_init`, `EMBEDDED_SWIFT_MUTEX_NUM_WORDS`, `swift_getPlatformLayerVersion`, etc.) against the current upstream `swiftlang/swift` source.
- [x] Confirmed the `<doc:STM32BaremetalGuide>` cross-reference resolves under main's `Examples/` layout.
- [ ] Build the DocC bundle locally and confirm the new article renders and appears under SDK support.

### DIFF
--- a/Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
+++ b/Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
@@ -27,7 +27,7 @@ void _swift_deallocate(void *ptr, size_t alignment, size_t size, swift_dealloc_f
 }
 ```
 
-### Provide console output
+## Provide console output
 
 `_swift_writeToStandardOutput(chars:count:)` is only needed if you use `print()` or `debugPrint()`. It receives a buffer of UTF-8 code points (not null-terminated) and should return the number of bytes written.
 
@@ -38,21 +38,21 @@ size_t _swift_writeToStandardOutput(const unsigned char *chars, size_t count) {
 }
 ```
 
-### Provide randomness
+## Provide randomness
 
 `_swift_generateRandom` feeds `SystemRandomNumberGenerator`, the default source used by `shuffle()` and friends. `_swift_generateRandomHashSeed` seeds the hashing used by `Set` and `Dictionary`; it doesn't need to be cryptographically secure, and can even return a fixed value if you want deterministic hashing. Both can typically forward to a hardware RNG or `arc4random_buf` where available.
 
-### Provide mutexes
+## Provide mutexes
 
 `_swift_mutex_init`, `_swift_mutex_destroy`, `_swift_mutex_lock`, `_swift_mutex_unlock`, and `_swift_mutex_tryLock` are only needed if you use `Synchronization.Mutex`. The caller allocates the storage for you — at least `EMBEDDED_SWIFT_MUTEX_NUM_WORDS` pointer-sized words (8 by default; override with `-Xcc -DEMBEDDED_SWIFT_MUTEX_NUM_WORDS=<n>` if your mutex representation needs more) — and hands it to `_swift_mutex_init` along with `.checked` and/or `.recursive` flags to opt into misuse diagnostics and reentrant locking.
 
 If your program is single-threaded and never preempted, you don't need a real lock at all: it's enough to track whether the mutex is currently held and trap on misuse, which is exactly what the standard library's single-threaded shim does. If you do have real concurrency, implement these on top of whatever your platform provides — a hardware spinlock, an RTOS mutex, or `pthread_mutex_t` on a POSIX-like target.
 
-### Proivde exclusivity checking
+## Provide exclusivity checking
 
 `_swift_getExclusivityTLS` and `_swift_setExclusivityTLS` are only needed when the compiler is built with `-enforce-exclusivity=checked`. They store and retrieve a single pointer per thread of execution. On a single-threaded platform this is just a global variable; on a multi-threaded platform it needs real thread-local storage.
 
-### Provide exiting
+## Provide exiting
 
 `_swift_exit(code:)` terminates the program and must not return. On a hosted platform this is a direct call to `exit()`; on baremetal it's usually an infinite loop or a reset.
 
--- a/Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
+++ b/Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
@@ -1,0 +1,98 @@
+# Bringing Embedded Swift to a new platform
+
+Implement the small set of runtime entry points that Embedded Swift expects from any platform, hosted or baremetal.
+
+## Overview
+
+<doc:IntegratingWithPlatforms> and <doc:Baremetal> describe how to connect Swift code to an *existing* SDK, or how to write the startup code and register access for a *specific* board. This article is about the layer underneath both of those: the small number of low-level functions that the Embedded Swift compiler and standard library call into whenever a program uses a feature that needs help from the outside world, such as allocating memory, calling `print()`, or creating a `Mutex`.
+
+A platform is anything that provides those functions: from a couple of hand-written functions in a single-file baremetal program, up to a full SDK or RTOS. Bringing Embedded Swift to a new platform means deciding, for each Swift feature you plan to use, how your target satisfies that feature's runtime dependency.
+
+Recent Swift toolchains formalize this contract as the Embedded Swift Platform Abstraction Layer, declared in  [EmbeddedPlatform.h](https://github.com/swiftlang/swift/blob/main/stdlib/public/EmbeddedPlatform/swift/EmbeddedPlatform.h). Every function declared in that header file matches the signature of a familiar POSIX function, so on a POSIX-like platform many of these can be one-line pass-throughs. None of them are required unconditionally. The linker only pulls in the ones your program actually exercises, so a program that never allocates on the heap doesn't need `_swift_allocate`, and a program that never creates a mutex doesn't need those functions.
+
+## Provide memory allocation
+
+`_swift_allocate(alignment:size:flags:)` and `_swift_deallocate(pointer:alignment:size:flags:)` back every heap allocation Embedded Swift performs implicitly: class instances, copy-on-write buffers for `Array`/`Dictionary`/`Set`/`String`, and explicit calls like `UnsafeMutablePointer.allocate`. A `SWIFT_ALLOC_ZERO_MEMORY` flag asks for zeroed memory, similar to `calloc`. `_swift_typedAllocate` is only needed if you enable the `TypedAllocation` feature; otherwise it can simply forward to `_swift_allocate`.
+
+```c
+void *_swift_allocate(size_t alignment, size_t size, swift_alloc_flags_t flags) {
+  void *ptr = NULL;
+  if (posix_memalign(&ptr, alignment, size) != 0) return NULL;
+  if (flags & SWIFT_ALLOC_ZERO_MEMORY) memset(ptr, 0, size);
+  return ptr;
+}
+
+void _swift_deallocate(void *ptr, size_t alignment, size_t size, swift_dealloc_flags_t flags) {
+  free(ptr);
+}
+```
+
+### Provide console output
+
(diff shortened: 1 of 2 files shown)


Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
```diff
@@ -0,0 +1,98 @@
+# Bringing Embedded Swift to a new platform
+
+Implement the small set of runtime entry points that Embedded Swift expects from any platform, hosted or baremetal.
+
+## Overview
+
+<doc:IntegratingWithPlatforms> and <doc:Baremetal> describe how to connect Swift code to an *existing* SDK, or how to write the startup code and register access for a *specific* board. This article is about the layer underneath both of those: the small number of low-level functions that the Embedded Swift compiler and standard library call into whenever a program uses a feature that needs help from the outside world, such as allocating memory, calling `print()`, or creating a `Mutex`.
```
heckj: When I read through this, I wasn't sure if this meant the article was about doing things that get you to an SDK (prerequisites) and there's more to building an SDK than is detailed here, or if this porting guide is effectively "how to create a beta SDK for your personal use" kind of thing.

Some of this is my own lack of understanding of what you need to use an arbitrary board - do you need an SDK, or is that more of a "hardened convenience" around these bare "gotta have" technical capabilities?

heckj (commented):
Looks great to me overall - Only real feedback was that it wasn't clear to me what a user needs to do with the implementations and how to wrap that into something broader, but it's also not something I've tried to do before - so maybe it's more obvious to folks who are wrangling Embedded and board bring ups.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DrPitre: It's meant for people bringing Embedded Swift to a new platform (e.g. microprocessor or operating system environment). Granted, it something that probably doesn't happen often, but having an article like this demonstrates that we *want* people to consider expanding the Embedded Swift ecosystem.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
```diff
@@ -0,0 +1,98 @@
...
+}
+```
+
+## Provide randomness
+
+`_swift_generateRandom` feeds `SystemRandomNumberGenerator`, the default source used by `shuffle()` and friends. `_swift_generateRandomHashSeed` seeds the hashing used by `Set` and `Dictionary`; it doesn't need to be cryptographically secure, and can even return a fixed value if you want deterministic hashing. Both can typically forward to a hardware RNG or `arc4random_buf` where available.
+
+## Provide mutexes
+
+`_swift_mutex_init`, `_swift_mutex_destroy`, `_swift_mutex_lock`, `_swift_mutex_unlock`, and `_swift_mutex_tryLock` are only needed if you use `Synchronization.Mutex`. The caller allocates the storage for you — at least `EMBEDDED_SWIFT_MUTEX_NUM_WORDS` pointer-sized words (8 by default; override with `-Xcc -DEMBEDDED_SWIFT_MUTEX_NUM_WORDS=<n>` if your mutex representation needs more) — and hands it to `_swift_mutex_init` along with `.checked` and/or `.recursive` flags to opt into misuse diagnostics and reentrant locking.
```
gonzalolarralde: A small update, [this PR](https://github.com/swiftlang/swift/pull/90960/changes) moved back to having separate function definitions for recursive threads. 

```
void _swift_mutexRecursive_init(void * EMBEDDED_SWIFT_NONNULL mutex,
                                swift_mutex_flags_t flags);
void _swift_mutexRecursive_destroy(void * EMBEDDED_SWIFT_NONNULL mutex);
void _swift_mutexRecursive_lock(void * EMBEDDED_SWIFT_NONNULL mutex);
void _swift_mutexRecursive_unlock(void * EMBEDDED_SWIFT_NONNULL mutex);
```

FWIW, `Synchronization.Mutex` doesn’t expose recursive locking. The recursive mutex functions are currently used internally by Concurrency, so perhaps they can be left out of this section and introduced later when discussing Concurrency support.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
```diff
@@ -0,0 +1,98 @@
...
+```c
+size_t _swift_writeToStandardOutput(const unsigned char *chars, size_t count) {
+  for (size_t i = 0; i < count; i++) putchar(chars[i]);
+  return count;
+}
+```
+
+## Provide randomness
+
+`_swift_generateRandom` feeds `SystemRandomNumberGenerator`, the default source used by `shuffle()` and friends. `_swift_generateRandomHashSeed` seeds the hashing used by `Set` and `Dictionary`; it doesn't need to be cryptographically secure, and can even return a fixed value if you want deterministic hashing. Both can typically forward to a hardware RNG or `arc4random_buf` where available.
```
DougGregor: `_swift_generateRandom` is only needed if you use `SystemRandomNumberGenerator`. It might be worth calling that out explicitly. I also recommend splitting this into two paragraphs: one for this entrypoint, and the other for hashing, because they're different uses.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DougGregor: Personally, I like documenting both recursive and non-recursive in the same place still, with a note that it's the non-recursive one needed for `Synchronization`.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
```diff
@@ -0,0 +1,98 @@
...
+
+If your program is single-threaded and never preempted, you don't need a real lock at all: it's enough to track whether the mutex is currently held and trap on misuse, which is exactly what the standard library's single-threaded shim does. If you do have real concurrency, implement these on top of whatever your platform provides — a hardware spinlock, an RTOS mutex, or `pthread_mutex_t` on a POSIX-like target.
+
+## Provide exclusivity checking
+
+`_swift_getExclusivityTLS` and `_swift_setExclusivityTLS` are only needed when the compiler is built with `-enforce-exclusivity=checked`. They store and retrieve a single pointer per thread of execution. On a single-threaded platform this is just a global variable; on a multi-threaded platform it needs real thread-local storage.
```
DougGregor: These entrypoints have been subsumed by `_swift_tls_*` entrypoints.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Provide exiting
+
+`_swift_exit(code:)` terminates the program and must not return. On a hosted platform this is a direct call to `exit()`; on baremetal it's usually an infinite loop or a reset.
+
+## Implement entry points in C or in Swift
+
+You can implement any of these functions in C. Define a function matching the declaration in `EmbeddedPlatform.h` and link it into your firmware, the same way you would provide any other C symbol the linker asks for.
```
DougGregor: I feel like this section belongs toward the beginning, because it lays out how you implement the functions that are described by most of the document.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
```diff
@@ -0,0 +1,98 @@
...
+  while true {}
+}
+```
+
+`@export(interface)` gives the function a single, externally visible definition, which is what lets the separately-compiled standard library link against it by symbol name.
+
+## Reuse the built-in shims
+
+Because every entry point is just a function with a specified signature, you rarely need to write all of them from scratch. The Swift repository ships several ready-made shims under [`stdlib/public/EmbeddedPlatform`](https://github.com/swiftlang/swift/tree/main/stdlib/public/EmbeddedPlatform) that you can use directly (where your toolchain provides prebuilt libraries for your target) or copy as a starting point:
+
```
DougGregor: I feel like this part about the built-in shims belongs way at the beginning, before we discuss how to implement your own. If you're doing Embedded Swift on a POSIX-y platform. If you're single-threaded,  against the provided swiftEmbeddedPlatformPOSIX and then choose between the single- and multi-threaded versions. We can show some link lines and paths to where to find them in the toolchain, because for some folks, this is all you need.

It's only if you don't have a POSIX-y platform that you need the rest of this document.

DougGregor (commented):
Thank you for writing this! I have a couple of "flow" comments, and we've tuned a few things in the platform abstraction layer since you started.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DrPitre: I verified against the current EmbeddedPlatform.h and confirmed `_swift_mutexRecursive_init`/`destroy`/`lock`/`unlock` are back as a separate function family with their own `EMBEDDED_SWIFT_MUTEX_RECURSIVE_NUM_WORDS` sizing. So I removed the incorrect `.recursive` flag and documented both families.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DrPitre: I split into two paragraphs: one for `_swift_generateRandom` with an explicit call-out that it's only needed for  `SystemRandomNumberGenerator`, and a separate one for `_swift_generateRandomHashSeed` covering `Set`/`Dictionary` hashing.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DrPitre: I went with documenting both families in the same "Provide mutexes" section, as you suggested, with non-recursive first with a note that `Synchronization.Mutex` only needs that family, then the recursive family with a note that it's currently only used internally by `Concurrency`.

I replaced the exclusivity section with "Provide thread-local storage," covering the current `_swift_tls_init`/`_swift_tls_get`/`_swift_tls_set` entry points, `SWIFT_TLS_KEY_COUNT`, and `_swift_thread_isMain`

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DrPitre: I moved this section up, right after "Reuse the built-in shims" and before the per-feature breakdown.

Sources/EmbeddedSwift/EmbeddedSwift.docc/SDKSupport/PortingToNewPlatforms.md
DrPitre: Ok. I moved this section to the front, right after the Overview, and reworded it to lead with the POSIX case: link  `swiftEmbeddedPlatformPOSIX` plus a threading shim (single- or multi-threaded) and you're generally done. I also added the actual library product names (`swiftEmbeddedPlatformPOSIX`, etc.) and an example link-line so it's actionable without reading the rest of the article. The per-feature sections that follow are now clearly for platforms that don't fit one of these shims.

DougGregor (approved):
Thank you for the revisions, this looks good to me!

